PR #21, opened 2021-06-04
ci: Create semantic-pull-request.yml

### LINKED ISSUE #5: CI: Add semantic PR GH workflow
Similar to [what we have set up](https://github.com/tendermint/starport/blob/develop/.github/workflows/semantic-pull-request.yml) for `tendermint/starport`, so that it's easier to read git history and write changelogs.

wal33d006:
Test PR for semantic workflow

Fixes #5


fadeev (approved):
✅

PR #29, opened 2021-06-15
feat: Mnemonic generation

wal33d006:
- Add mnemonic generation for creating a new wallet
- Add encryption technique
- Storing Mnemonic in secure storage
- Now if the user adds and configures the wallet, and closes the app. After he opens the app again, he will be asked for a password to enter into the app to see his wallet.

The flow of storing Mnemonic when user enters the app for the first time:
- Generate a mnemonic and show it to the user
- Ask user for a password
- Encrypt the mnemonic and store it inside secure storage
- Add a bool variable in shared preferences in order to know if the user has already configured the wallet
- Import the wallet through mnemonic and take the user to wallet listing page

Flow when user has already configured the wallet and comes back to the app:
- Get the bool from shared preferences
- If the bool is not null, take the user to password page
- User enters the password which is used to decrypt the mnemonic which was stored previously in secure storage
- If the mnemonic is decrypted, the user is taken to the wallet listing page.

Simulator Screen Shot - iPhone 12 Pro - 2021-06-15 at 11 55 28
Simulator Screen Shot - iPhone 12 Pro - 2021-06-15 at 11 55 31
Simulator Screen Shot - iPhone 12 Pro - 2021-06-15 at 11 55 35


wal33d006:
Updated the encryption decryption logic using stretch to produce a key of desired length of 24 characters to encrypt the mnemonic

lib/api_calls/faucet_api.dart
```diff
@@ -0,0 +1,18 @@
+import 'dart:convert';
+
+import '../global.dart';
+
+class FaucetApi {
+  getFreeTokens(String address) async {
```
andrzejchm, line 6: we should add return types to all methods, here `void`. I'll also make sure to add some linter later on

lib/global.dart
```diff
@@ -35,3 +35,11 @@ class BaseEnv {
 }
 
 final BaseEnv baseEnv = BaseEnv();
+
+
+
+abstract class SharedPreferencesKeys {
+  static const isWalletCreated = 'IS_APP_INSTALLED';
```
andrzejchm, line 42: we're using those keys in a single class just, so we shouldn't need a global class with keys. Later on we will have abstraction arround shared prefs and secure storage, which will encapsulate the way we interact with those and keys can be private and local to that class only

lib/helpers/mnemonic_encryptor.dart
```diff
@@ -0,0 +1,61 @@
+import 'dart:typed_data';
+
+import 'package:encrypt/encrypt.dart';
+import 'package:flutter/foundation.dart' as foundation;
+import 'package:flutter_secure_storage/flutter_secure_storage.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+import '../global.dart';
+
+abstract class MnemonicEncryptor {
```
andrzejchm, line 10: we'll need to refactor this a bit in order to support testability and dependency injection, but that's after we come up with a common architecture for the mobile app

andrzejchm (changes requested):

andrzejchm:
This looks very solid and a good amount of thought! Thanks @wal33d006 !!

lib/global.dart
wal33d006: This key is used in two different classes `mnemonic_encryptor.dart` and `mnemonic_onboarding.dart`. I just created this small class so that we don't need to maintain the key in two different places. We can refactor this when we place an abstraction around this. Let me know how to use these keys in two classes if I remove this class from here. Should I use them directly as strings?

andrzejchm: that's absolutely fine then! I've wrongly assumed it's used currently in one place. We'll deal with that when we have abstraction arround shared prefs

lib/api_calls/faucet_api.dart
wal33d006: Updated all functions to have return types 🙌

andrzejchm (approved):

alain2sf:
> Updated the encryption decryption logic using stretch to produce a key of desired length of 24 characters to encrypt the mnemonic

We should use a 32 bytes (256 bits) key space, as the AES scheme is applied, therefore getting AES-256 level. The AES block being 128 bits it will anyways be more efficient and require less padding operations.

wal33d006:
> > Updated the encryption decryption logic using stretch to produce a key of desired length of 24 characters to encrypt the mnemonic
> 
> We should use a 32 bytes (256 bits) key space, as the AES scheme is applied, therefore getting AES-256 level. The AES block being 128 bits it will anyways be more efficient and require less padding operations.

Just did that in bb605a5

fadeev (approved):
Tested on the web, looks good 👍

PR #43, opened 2021-06-22
feat: Add UI support for Ethereum

### LINKED ISSUE #32: feat: Add UI support for ETH based wallets

wal33d006:
Fixes #32 
Simulator Screen Shot - iPhone 12 Pro - 2021-06-21 at 17 07 58
Simulator Screen Shot - iPhone 12 Pro - 2021-06-21 at 17 09 18


lib/api_calls/eth_api.dart
```diff
@@ -0,0 +1,69 @@
+import 'dart:math';
+import 'package:flutter_app/api_calls/base_wallet_api.dart';
+import 'package:flutter_app/global.dart';
+import 'package:flutter_app/models/balances.dart';
+import 'package:flutter_app/models/eth_wallet.dart';
+import 'package:http/http.dart';
+import 'package:wallet_core/wallet_core.dart';
+
+class EthApi extends BaseWalletApi {
```
andrzejchm, line 9: could we rename it to `EthereumApi`?

lib/global.dart
```diff
@@ -27,15 +32,23 @@ class BaseEnv {
...
   String get baseApiUrl => _baseApiUrl;
+
+  String get baseEthUrl => _baseEthUrl;
 }
 
 final BaseEnv baseEnv = BaseEnv();
 
+final CosmosApi cosmosApi = CosmosApi();
+final EthApi ethApi = EthApi();
+
```
andrzejchm, line 49: lets remove the extra blank lines

lib/models/cosmos_wallet.dart
```diff
@@ -0,0 +1,30 @@
...
+  String walletAddress;
+  String walletAlias;
+  WalletType walletType;
+
+  BaseWalletDetails({
+    required this.walletType,
+    required this.walletAlias,
+    required this.walletAddress,
+  });
+}
```
andrzejchm, line 28: lets move it to separate file. Also, we can just name it `WalletDetails`.

lib/models/cosmos_wallet.dart
```diff
@@ -0,0 +1,30 @@
...
+    required this.wallet,
+  }) : super(
+          walletAlias: walletAlias,
+          walletAddress: walletAddress,
+          walletType: walletType,
+        );
+}
+
+class BaseWalletDetails {
+  String walletAddress;
```
andrzejchm, line 19: all the classes that are merely holding data should be immutable as much as possible, thus all the fields should be marked as final. we can then also mark the constructor with `const` keyword.

We can also rename the class to `WalletDetails`, no need for the `Base` prefix

lib/models/cosmos_wallet.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:sacco/wallet.dart';
+
+class CosmosWallet extends BaseWalletDetails {
```
andrzejchm, line 3: we should avoid inheritance wherever possible and stick to composition. We could simply get rid of the `CosmosWallet` and `EthWallet` classes as they don't bring any additional value here

lib/models/cosmos_wallet.dart
```diff
@@ -0,0 +1,30 @@
...
+  WalletType walletType;
+
+  BaseWalletDetails({
+    required this.walletType,
+    required this.walletAlias,
+    required this.walletAddress,
+  });
+}
+
+enum WalletType { Eth, Cosmos }
```
andrzejchm, line 30: this could be moved into separate file aswell

lib/views/password_generation.dart
```diff
@@ -75,11 +80,16 @@ class _PasswordGenerationPageState extends State<PasswordGenerationPage> {
   }
 
   Future<void> encryptMnemonic(BuildContext context) async {
-    await MnemonicEncryptor.encryptMnemonic(widget.mnemonic!, passwordController.text);
-    final api = WalletApi();
-    api.importWallet(
+    await MnemonicEncryptor.encryptMnemonic(
+        widget.mnemonic!, passwordController.text);
```
andrzejchm, line 84: I think you have the line length set to 80 chars instead of 120

test/transaction_test_local.dart
```diff
@@ -1,15 +1,19 @@
 import 'package:flutter/foundation.dart';
-import 'package:flutter_app/api_calls/wallet_api.dart';
+import 'package:flutter_app/api_calls/cosmos_api.dart';
 import 'package:flutter_app/global.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
```
andrzejchm, line 6: lets rename the file to `local_transaction_test.dart`

test/transaction_test_hosted.dart
```diff
@@ -1,18 +1,19 @@
 import 'package:flutter/foundation.dart';
+import 'package:flutter_app/api_calls/cosmos_api.dart';
 import 'package:flutter_app/api_calls/faucet_api.dart';
-import 'package:flutter_app/api_calls/wallet_api.dart';
 import 'package:flutter_app/global.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
```
andrzejchm, line 7: lets stick with a naming convetion that all test files are ending with `*_test.dart` suffix. this will be easier to distinguish the files while reviewing. here: `hosted_transaction_test.dart`

test/ethereum_test.dart
```diff
@@ -11,31 +11,37 @@ void main() {
...
       final ethClient = Web3Client(apiUrl, httpClient);
 
       final balance = await ethClient.getBalance(
         EthereumAddress.fromHex(wallet.privateKey.address.hex),
       );
+
+      debugPrint(wallet.privateKey.address.hex);
+      debugPrint(privateEthCredentials.address.hex);
       debugPrint("${balance.getValueInUnit(EtherUnit.ether)}");
-      await ethClient.sendTransaction(
```
andrzejchm, line 29: lets avoid commenting out the code, if we don't need it, just remove it. we can use git history in order to retrieve back the code if needed

macos/Flutter/Flutter-Debug.xcconfig
```diff
@@ -0,0 +1,2 @@
+#include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
```
andrzejchm, line 1: we don't need `macos` folder structure, at least for now, lets remove it

andrzejchm (changes requested):
solid work 💪🏻, I've left few comments for you to consider

lib/models/cosmos_wallet.dart
wal33d006: We are using this inheritance to cater generic wallets in our UI. We are also reusing code from `BaseWalletDetails` in order to keep things under abstraction for the UI.

lib/models/cosmos_wallet.dart
wal33d006: Prefix is used because it is the parent class of different types of wallets. Cosmos and Ethereum wallets extend this class.

lib/api_calls/eth_api.dart
wal33d006: ✅

lib/global.dart
wal33d006: ✅

lib/views/password_generation.dart
wal33d006: ✅

test/transaction_test_local.dart
wal33d006: ✅

test/transaction_test_hosted.dart
wal33d006: ✅

test/ethereum_test.dart
wal33d006: ✅

macos/Flutter/Flutter-Debug.xcconfig
wal33d006: ✅

lib/models/cosmos_wallet.dart
wal33d006: ✅

lib/models/cosmos_wallet.dart
wal33d006: ✅

lib/models/cosmos_wallet.dart
andrzejchm: you can leverage composition in following way:

```dart
 class CosmosWallet {
  final WalletDetails walletDetails;
  final Wallet wallet;
  final WalletType walletType;

  const CosmosWallet({
    required this.walletDetails,
    required this.wallet,
  }) : walletType = WalletType.cosmos;
}

...

class WalletDetails {
  final String walletAddress;
  final String walletAlias;

  const WalletDetails({
    required this.walletAddress,
    required this.walletAlias,
  });


}
```

I just have some problems with current naming, there's a lot of wallets right now, but we can work that out in a bit, when the final architecture outline is ready

wal33d006: Two confusions:
- In that case, how will the UI render different kinds of wallets altogether? The object will be of different types then. Currently the UI list only knows that there are wallets and not types of wallets.
- Since `CosmosWallet` is a `Wallet` and not a part of a wallet so composition in this case seems a little off to me. What do you think?

andrzejchm: in order to support multiple wallets then we can hide it behind abstraction like this:
```
abstract class EmerisWallet {
  WalletDetails get walletDetails;
  WalletType get walletType;
}

class CosmosWallet implements EmerisWallet {
    ...
}
```

Regarding the `Wallet` object from Sacco library, this should not be part of the domain entity and should not be accessed directly from the UI, since it is a specific implementation, that could also be hidden behind abstraction to allow for easy swapping in the future, but this can be worked on later on, in a separate pr.

wal33d006: This sounds cool. Yes the Wallet object from libraries need to be abstracted, they should not be accessed by the UI, which is why we will fix that in the architecture. I am, however, still not convinced about using composition instead of inheritance. In real world scenarios, it's about being a part of. In our case it's otherwise.

andrzejchm: With composition we allow for better code reusability and readability. if we stick to inheritance, we can end up in inheritance hell with multiple levels of `extends`. this makes hard to determine what hte code is actually doing without having to navigate trough multiple levels of classes. Here the `EmerisWallet` consists of `WalletDetails` and `WalletType` which defines that each wallet in our app should support having details of the wallet `i.e: public address and alias` as well as `walletType`

wal33d006: Perfect. Now it makes sense. The `WalletDetails` is now not a part of `CosmosWallet`, it's inside the abstraction now. Am I getting it right now?

andrzejchm: yes, pretty much, though this abstraction is merely an interface, thus CosmosWallet , when implementing the abstraction will need to expose a walletDetails field

wal33d006: Updated ✅

lib/models/cosmos_wallet.dart
```diff
@@ -0,0 +1,19 @@
...
+    required this.baseWalletDetails,
+    required this.wallet,
+  });
+
+  @override
+  WalletDetails get walletDetails => baseWalletDetails;
+
+  @override
+  WalletType get walletType => WalletType.Cosmos;
+}
```
andrzejchm, line 19: you can simplify it a bit:
```suggestion
class CosmosWallet implements EmerisWallet {
  final Wallet wallet;
  @override
  final WalletDetails walletDetails;

  CosmosWallet({
    required this.walletDetails,
    required this.wallet,
  });

  @override
  WalletType get walletType => WalletType.Cosmos;
}
```

lib/models/emeris_wallet.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:flutter_app/models/wallet_type.dart';
+
+class WalletDetails {
```
andrzejchm, line 3: we should put each class in separate file for clarity

lib/models/cosmos_wallet.dart
wal33d006: Done for both Ethereum and Cosmos wallets ✅

andrzejchm (approved):
🚀

PR #46, opened 2021-06-25
feat: Add Intellij's File templates

andrzejchm:
Those are templates for MVP (model-view-presenter) page structure as well as use case's template


wal33d006 (approved):
🚀

PR #51, opened 2021-06-28
feat: Firebase app distribution

### LINKED ISSUE #52: Firebase App Distribution setup
- on each commit to open pr we build and distribute android and iOS version to testers and add a comment to the pr with a download link. Only approved testers can follow it, while others will receive 404. there should be a separate testers group on firebase set up for that

- on each commit to `main` branch, distribute app version meant for broader audience, interrested only in a working version of the app. there should be a separate testers group on firebase set up for that.

Each of those versions should have a changelog set up out of the last 5 commits

andrzejchm:
fixes #52

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -2,14 +2,16 @@ name: Pull Request Worfklow
 
 on: [pull_request]
 
+env:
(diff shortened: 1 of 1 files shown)


andrzejchm:
will be handled in private repo

PR #53, opened 2021-06-28
feat: Localization support

### LINKED ISSUE #48: Add Localizations support to the app
use the tutorial here: https://flutter.dev/docs/development/accessibility-and-localization/internationalization

to set up `intl` package that will generate localized strings based on the `arb` files. For now lets add only the `en` language

wal33d006:
Fixes #48


lib/views/password_generation.dart
```diff
@@ -24,7 +25,7 @@ class _PasswordGenerationPageState extends State<PasswordGenerationPage> {
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
-        title: const Text('Password generation'),
+        title: Text(AppLocalizations.of(context)!.appTitle),
```
andrzejchm, line 28: we should avoid force unwrap wherever possible.  In this case `AppLocalizations.of(context)?.appTitle` should be enough

andrzejchm (commented):
Great job! 🥇 I'm thinking about one improvement we can add here:

Since there will be plenty of places where we will want to use strings, where `context` is not available, we could create a static field to access it without context, for example:

in main.dart:
```dart
class EmerisApp extends StatefulWidget {
  @override
  _EmerisAppState createState() => _EmerisAppState();
}

class _EmerisAppState extends State<EmerisApp> {
  @override
  void initState() {
    super.initState();
   initializeAppLocalizations(context);
  }

  @override
    return const AppWidget();
  }
}
```
and add new file: in `lib/utils/strings.dart`:
```dart
/// Convenience static field to hold the reference to AppLocalizations object for easy access without BuildContext
static AppLocalizations strings;

void initializeAppLocalizations(BuildContext context) {
    strings = AppLocalizations.of(context) ?? (throw StateError("Could not initialize App Localizations"));
}
```

lib/views/password_generation.dart
wal33d006: ✅

wal33d006:
> Great job! 🥇 I'm thinking about one improvement we can add here:
> 
> Since there will be plenty of places where we will want to use strings, where `context` is not available, we could create a static field to access it without context, for example:
> 
> in main.dart:
> 
> ```dart
> class EmerisApp extends StatefulWidget {
>   @override
>   _EmerisAppState createState() => _EmerisAppState();
> }
> 
> class _EmerisAppState extends State<EmerisApp> {
>   @override
>   void initState() {
>     super.initState();
>    initializeAppLocalizations(context);
>   }
> 
>   @override
>   Widget build(BuildContext context) {
>     return const AppWidget();
>   }
> }
> ```
> 
> and add new file: in `lib/utils/strings.dart`:
> 
> ```dart
> /// Convenience static field to hold the reference to AppLocalizations object for easy access without BuildContext
> static AppLocalizations strings;
> 
> void initializeAppLocalizations(BuildContext context) {
>     strings = AppLocalizations.of(context) ?? (throw StateError("Could not initialize App Localizations"));
> }
> ```
> 
> then, on the call site we can do something like this:
> 
> ```dart
> Text(strings.appTitle);
> ```

✅

andrzejchm (approved):
🚀

PR #59, opened 2021-06-29
feat: Introduce routing page

andrzejchm:
Routing page is used to initialize app and determine which page should be shown at start


andrzejchm:
Will be handled in private repo

PR #62, opened 2021-07-06
feat: key info storage implementation

andrzejchm:


packages/transaction_signing_gateway/lib/encrypt/aes_cipher.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:encrypt/encrypt.dart';
+import 'package:transaction_signing_gateway/encrypt/cipher.dart';
+
+class AESCipher implements Cipher {
```
andrzejchm, line 4: @alain2sf  can you take a look if this implementation of encrypting/decrypting data using just password makes sense?

packages/transaction_signing_gateway/test/aes_cipher_test.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:transaction_signing_gateway/encrypt/aes_cipher.dart';
+
+void main() {
```
andrzejchm, line 4: @alain2sf  here are some tests to prove encrypting/decrypting works and that we use random salt on each encryption

packages/transaction_signing_gateway/lib/encrypt/aes_cipher.dart
```diff
@@ -0,0 +1,56 @@
...
+      throw ArgumentError("encrypted data has wrong format");
+    }
+    return encrypter.decrypt(
+      Encrypted.fromBase64(bundle.encryptedData),
+      iv: IV.fromBase64(bundle.iv),
+    );
+  }
+
+  @override
+  String encrypt({required String password, required String data}) {
```
andrzejchm, line 19: updated the code, now when encrypting we're generating random, 16bytes long IV, use it to encrypt the data and prepend it to the encrypted string.

packages/transaction_signing_gateway/lib/encrypt/aes_cipher.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:encrypt/encrypt.dart';
+import 'package:transaction_signing_gateway/encrypt/cipher.dart';
+
+class AESCipher implements Cipher {
+  @override
+  String decrypt({required String password, required String encryptedData}) {
+    final encrypter = _fromPassword(password);
```
andrzejchm, line 7: `encryptedData` consists of IV and encrypted message, we split that up with `_encryptedBundle.fromString(...)` and use the data to perform decryption

packages/transaction_signing_gateway/lib/encrypt/aes_cipher.dart
```diff
@@ -0,0 +1,56 @@
...
+  final String iv;
+  final String encryptedData;
+
+  _EncryptedBundle({
+    required this.iv,
+    required this.encryptedData,
+  });
+
+  static _EncryptedBundle? fromString(String str) {
+    final split = str.split(_delimiter);
```
andrzejchm, line 47: constructor that splits the input string into an iv and encryptedData, both are base64 encoded

packages/transaction_signing_gateway/lib/encrypt/aes_cipher.dart
```diff
@@ -0,0 +1,56 @@
...
+    if (split.length != 2) {
+      return null;
+    } else {
+      return _EncryptedBundle(iv: split[0], encryptedData: split[1]);
+    }
+  }
+
+  String get string => "$iv$_delimiter$encryptedData";
```
andrzejchm, line 55: this generates a single string that consists of IV and encrypted data. delimited with `|` sign

wal33d006 (approved):
🚀

packages/transaction_signing_gateway/lib/encrypt/aes_cipher.dart
alain2sf: Again I'm not familiar with Dart language, hence not too comfortable reading the code. It seems the IV is generated from the password, we should generate the IV as a random nonce of 128 bits. And use a new unique random IV for each encryption. Salt is 256 bits long? Can't figure that out. 
Salt and IV should not be in Base64 format when used in encryption/derivation techniques, it decreases the entropy. 
See my message in Slack #dev-mobile where I describe the encryption/decryption flows (pseudo code) with the recommendations. Thanks!

alain2sf (changes requested):

packages/transaction_signing_gateway/lib/encrypt/aes_cipher.dart
andrzejchm: currently, IV is generated for each encryption using SecureRandom, 128bit long (16 bytes long), password is not taken into consideration when generating it.
```dart
IV _iv() => IV.fromSecureRandom(16);
```
Salt is also generated from SecureRandom as 32bytes long array:
```dart
IV _salt() => IV.fromSecureRandom(32);
```

I've updated the code to do base64 encryption only as the last step before spitting out the encryption result. We base64 encode IV and salt to be able to represent them as strings. they all get concatenated together into the final result of encryption so that we can decrypt the message afterwards. i.e:

```dart
"${base64(IV)} | ${base64(Salt)} | ${base64(EncryptedMessage)}"
````

andrzejchm:
@alain2sf this is now ready for re-review. I've updated the code so that base64 encoding is done only when spitting out final result for encryption and when reading input for decryption

andrzejchm:
@alain2sf do you mind re-reviewing the pr after the changes? :)

PR #72, opened 2021-08-02
feat: Transaction broadcasting

wal33d006:


packages/transaction_signing_gateway/lib/alan/alan_transaction_broadcaster.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'package:transaction_signing_gateway/model/transaction_broadcasting_failure.dart';
+import 'package:transaction_signing_gateway/model/transaction_hash.dart';
+import 'package:transaction_signing_gateway/transaction_broadcaster.dart';
+
+class AlanTransactionBroadcaster implements TransactionBroadcaster {
+  @override
+  Future<Either<TransactionBroadcastingFailure, TransactionHash>> broadcast(
+      {required SignedTransaction transaction, required PrivateWalletCredentials privateWalletCredentials}) async {
+    if (transaction is! SignedAlanTransaction) {
+      return left(AlanTransactionBroadcastingFailure('passed transaction is not $UnsignedAlanTransaction'));
```
andrzejchm, line 16:
```suggestion
      return left(AlanTransactionBroadcastingFailure('passed transaction is not $SignedAlanTransaction'));
```

packages/transaction_signing_gateway/lib/model/transaction_broadcasting_failure.dart
```diff
@@ -0,0 +1,5 @@
+abstract class TransactionBroadcastingFailure {}
```
andrzejchm, line 1: we'd need a `type` here, if possible. similar to what we have in `TransactionSigningFailure` in order to allow end users to `switch` over possible broadcasting errors.

andrzejchm (changes requested):
two small things  :)

packages/transaction_signing_gateway/lib/alan/alan_transaction_broadcaster.dart
wal33d006: ✅

packages/transaction_signing_gateway/lib/model/transaction_broadcasting_failure.dart
wal33d006: ✅

andrzejchm (approved):
🚀 let's 🚢 it!

PR #85, opened 2021-08-16
refactor: Add version for `alan` in pubspec.yaml

wal33d006:


starport_template/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ dependencies:
     sdk: flutter
 
   mobx: 2.0.3
-  alan:
+  alan: 0.40.2+1
```
andrzejchm, line 14: you can simply remove this line completely, since `transaction_signing_gateway` is depending on alan and the dependency will become transitive. this way we make sure both template and `transaciton_signing_gateway` use exactly the same version

wal33d006: ✅

andrzejchm (approved):

PR #88, opened 2021-08-17
refactor: Wallet derivation

### LINKED ISSUE #87: refactor: Add wallet derivation in `transaction_signing_gateway`

wal33d006:
Fixes #87


packages/transaction_signing_gateway/lib/wallet_derivator.dart
```diff
@@ -0,0 +1,12 @@
...
+import 'package:transaction_signing_gateway/model/network_info.dart';
+import 'package:transaction_signing_gateway/model/private_wallet_credentials.dart';
+import 'package:transaction_signing_gateway/model/wallet_derivation_failure.dart';
+
+abstract class WalletDerivator {
+  Either<WalletDerivationFailure, PrivateWalletCredentials> derive({
+    required String mnemonic,
+    required String walletAlias,
+    required NetworkInfo networkInfo,
+  });
```
andrzejchm, line 11: we cannot accept NetworkInfo here, as this model is specific to Alan and Cosmos network. what we can do instead is to accept following:

```suggestion
  Either<WalletDerivationFailure, PrivateWalletCredentials> derive({
    required WalletDerivationInfo walletDerivationInfo
  });
```
Add new interface for `WalletDerivationInfo`:

```dart
abstract class WalletDerivationInfo {
}
```

and implementation specific for Alan

```dart
class AlanWalletDerivationInfo implements WalletDerivationInfo {
    final String mnemonic;
    final String walletAlias;
    final NetworkInfo networkInfo; // Alan's version of network info, we don't need a separate model in the transaction_signing_gateway library
    
}
```

packages/transaction_signing_gateway/lib/model/network_info.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:equatable/equatable.dart';
+
+class NetworkInfo extends Equatable {
```
andrzejchm, line 3: we can remove that after suggested changes are in place

packages/transaction_signing_gateway/lib/wallet_derivator.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:dartz/dartz.dart';
+import 'package:transaction_signing_gateway/model/network_info.dart';
+import 'package:transaction_signing_gateway/model/private_wallet_credentials.dart';
+import 'package:transaction_signing_gateway/model/wallet_derivation_failure.dart';
+
+abstract class WalletDerivator {
```
andrzejchm, line 6: We would also need to add a `derive` method inside the `TransactionSigningGateway` class and make it work similar to what we have for transaction signing, where we have a list of derivators and look for one which is capable of handling the derivation

andrzejchm (changes requested):

packages/transaction_signing_gateway/lib/wallet_derivator.dart
wal33d006: ✅

packages/transaction_signing_gateway/lib/model/network_info.dart
wal33d006: ✅

packages/transaction_signing_gateway/lib/wallet_derivator.dart
wal33d006: ✅

andrzejchm (changes requested):

packages/transaction_signing_gateway/lib/gateway/transaction_signing_gateway.dart
```diff
@@ -78,6 +85,10 @@ class TransactionSigningGateway {
                 privateWalletCredentials: privateCreds,
               ));
 
+  Future<Either<WalletDerivationFailure, PrivateWalletCredentials>> deriveWallet(
+          {required AlanWalletDerivationInfo alanWalletDerivationInfo}) async =>
```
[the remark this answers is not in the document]
wal33d006: ✅

andrzejchm (approved):
🚀

PR #89, opened 2021-08-17
fix: MobX mutation actions

andrzejchm:
We have to wrap the observable mutations into actions in order to not cause the `Side effects like changing state are not allowed at this point` type of errors


wal33d006 (approved):
Looks great!

PR #92, opened 2021-08-20
refactor: User `transaction_signing_gateway` for wallet derivation

wal33d006:


andrzejchm:
I've made the same tweak on my branch along with few others, could we merge mine instead in order to avoid merge conflicts? I'll post a pr in a moment

PR #100, opened 2021-10-08
refactor: Add different font sizes

wal33d006:


packages/cosmos_ui_components/lib/cosmos_app_theme.dart
```diff
@@ -20,6 +20,9 @@ class CosmosAppTheme {
   static const radiusS = 4.0;
 
   static const fontSizeS = 12.0;
+  static const fontSizeM = 20.0;
+  static const fontSizeL = 28.0;
+  static const fontSizeXL = 40.0;
```
andrzejchm, line 25: I'm affraid the steps between S, M, L and XL are slightly too big, maybe:
```suggestion
  static const fontSizeS = 12.0;
  static const fontSizeM = 16.0;
  static const fontSizeL = 20.0;
  static const fontSizeXL = 28.0;
    static const fontSizeXXL = 40.0;
```

PR #101, opened 2021-10-08
style: Fix styling for CosmosElevatedButton and CosmosOutlinedButton

wal33d006:


andrzejchm:
for Outline Button you have the `CosmosOutlineButton` class, maybe you could leverage that?

packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
```diff
@@ -5,18 +5,24 @@ class CosmosElevatedButton extends StatelessWidget {
...
+    this.outlined = false,
   }) : super(key: key);
 
   @override
   Widget build(BuildContext context) {
     return ElevatedButton(
       onPressed: onTap,
+      style: ElevatedButton.styleFrom(
+        primary: Theme.of(context).colorScheme.secondary,
+        onPrimary: Theme.of(context).colorScheme.onSecondary,
```
wal33d006, line 24: @andrzejchm 
Removed the outline thing. Just keeping this color thing because the button was acting weird when switching between dark/light modes.

packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
```diff
@@ -5,18 +5,24 @@ class CosmosElevatedButton extends StatelessWidget {
   final VoidCallback? onTap;
   final String text;
   final Widget? suffixIcon;
+  final bool outlined;
```
andrzejchm, line 8: lets remove this param

packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
andrzejchm: Instead of playing with colors here, I think its better to update the cosmos theme accordingly in `CosmosAppTheme` class. right now we're setting secondary color to primary, which might cause some headscratching when someone decides to alter theme of their apps

packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
wal33d006: ✅

PR #110, opened 2021-11-01
refactor: Amount formatter

wal33d006:


packages/cosmos_utils/lib/amount_formatter.dart
```diff
@@ -0,0 +1,22 @@
...
+    }
+    if (raw.length >= 13) {
+      return _getShortAmount(raw, 12, 'T'); // Trillion
+    }
+    if (raw.length >= 10) {
+      return _getShortAmount(raw, 9, 'B'); // Billion
+    }
+    return _getShortAmount(raw, 6, 'M'); // Million
+  }
+  return formatted;
```
wal33d006, line 18: @andrzejchm 
Came up with this idea of adding `M`, `B`, `T`, and so on, for formatting amount along with simple formatting. Do let me know if we can think of something better than this.

andrzejchm: how about we use the `NumberFormat.compact()` formatter?

https://stackoverflow.com/a/54691084/4838100

wal33d006: That's perfect for our use case!

wal33d006: Updated ✅

packages/cosmos_utils/lib/amount_formatter.dart
```diff
@@ -0,0 +1,3 @@
+import 'package:intl/intl.dart';
+
+String formatAmount(double amount) => NumberFormat.compactCurrency(symbol: '\$').format(amount);
```
andrzejchm, line 3: let's make it a bit more sophisticated:
```suggestion
String formatAmount(double amount, {String locale = "en_US", String symbol = '\$'}) => amount >= 10000
    ? NumberFormat.compactCurrency(
        symbol: symbol,
        locale: locale,
      ).format(amount)
    : NumberFormat.currency(
        locale: locale,
        symbol: symbol,
      ).format(amount);

```

packages/cosmos_utils/test/number_formatter_test.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:cosmos_utils/amount_formatter.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  test('Amount formatter test case', () {
+    expect(formatAmount(37570000000), '\$37.6B');
+  });
```
andrzejchm, line 7:
```suggestion
    test('Amount formatter test case', () {
    expect(formatAmount(37570000000), '\$37.6B');
    expect(formatAmount(10270, locale: "pl", symbol: "PLN"), '10,3 tys. PLN');
    expect(formatAmount(1027.004921), '\$1,027.00');
    expect(formatAmount(10000), '\$10K');
    expect(formatAmount(9999), '\$9,999.00');
    expect(formatAmount(1027), '\$1,027.00');
  });
```

andrzejchm (changes requested):
I've added few more improvements so that the function is a bit more configurable

packages/cosmos_utils/test/number_formatter_test.dart
wal33d006: ✅

packages/cosmos_utils/lib/amount_formatter.dart
wal33d006: ✅

andrzejchm (approved):

PR #119, opened 2021-11-17
feat: Biometric storage [REVIEW AFTER #118]

andrzejchm:
This aims at adding new library `biometric_storage` which allows us to "guard" the wallet credentials data with biometric authentication. I've also refactored the approach to storages a bit, where general logic of storing and retrieving credentials data is contained within the `CosmosKeyInfoStorage`, while the implementation specifics of stores are hidden behind the abstractions:
- `PlainDataStore` - implemented by `SharedPrefsPlainDataStore`
- `SecureDataStore` - implemented by `BiometricDataStore` and `FlutterSecureStorageDataStore`


wal33d006 (approved):
Awesome!

PR #121, opened 2021-11-17
style: Add `gradient_avatar.dart` along with deterministic color generation from `walletAddress`

wal33d006:


wal33d006:
Simulator Screen Shot - iPhone 13 - 2021-11-17 at 15 54 50

packages/cosmos_ui_components/lib/components/gradient_avatar.dart
```diff
@@ -0,0 +1,41 @@
...
+
+  List<Color> generateColorsFromAddress(String address) {
+    final color1 = StringBuffer();
+    final color2 = StringBuffer();
+
+    color1.write('#');
+    color2.write('#');
+
+    // Initialised it here because linter was warning me not to declare multiple variables in single line
+    var j = address.length - 7;
```
andrzejchm, line 30: let's use more descriptive names for the variables. also the comment above is not needed, variables should always be initialized in separate lines for readability

packages/cosmos_ui_components/lib/components/gradient_avatar.dart
```diff
@@ -0,0 +1,41 @@
...
+      child: Container(
+        decoration: BoxDecoration(
+          shape: BoxShape.circle,
+          gradient: LinearGradient(colors: [colors[0], colors[1]]),
+        ),
+      ),
+    );
+  }
+
+  List<Color> generateColorsFromAddress(String address) {
```
andrzejchm, line 22: is this logic copied from web? we need to make sure we use the same logic as web version to stay consistent

packages/cosmos_ui_components/lib/components/gradient_avatar.dart
```diff
@@ -0,0 +1,41 @@
...
+    final color1 = StringBuffer();
+    final color2 = StringBuffer();
+
+    color1.write('#');
+    color2.write('#');
+
+    // Initialised it here because linter was warning me not to declare multiple variables in single line
+    var j = address.length - 7;
+    for (var i = address.length - 1; color1.length <= 7 && color2.length <= 7; i--, j--) {
```
andrzejchm, line 31: this is a bit strange construction of for loop with two separate iterators and condition not depending on the iterator itself. a `while` loop here would have more sense. also we're not handling the edge cases where the address might be shorter than 14 letters. how about we do something like:

```dart
    final color1 = address.substring(0, address.length ~/ 2 + 1);
    final color2 = address.substring(address.length ~/ 2 + 1);
    return [
      Color(StringToHex.toColor(color1)),
      Color(StringToHex.toColor(color2)),
    ];
```
but it will only make sense if web does the same thing.

packages/cosmos_ui_components/lib/components/gradient_avatar.dart
andrzejchm, line 22: also, we could extract this logic into `cosmos_utils` package, since it might come handy in multiple places

andrzejchm (changes requested):
Looks nice, lets consider reusability and parity with web though

packages/cosmos_ui_components/lib/components/gradient_avatar.dart
wal33d006: Not the same as web.

The web uses a package called [gradient avatar](https://www.npmjs.com/package/gradient-avatar) which actually does something like this inside it:
```
const hash = stringHash(str);
  const colors = hslTriad(hash % 360, 1, 0.5);
  const color1 = hslRgb(colors[0][0], colors[0][1], colors[0][2]);
  const color2 = hslRgb(colors[1][0], colors[1][1], colors[1][2]);
```

wal33d006: I think I can make something similar to web using `HSLColor` class in dart.

packages/cosmos_ui_components/pubspec.lock
```diff
@@ -68,8 +68,8 @@ packages:
     dependency: "direct main"
     description:
       path: "packages/cosmos_utils"
-      ref: starport-redesign
-      resolved-ref: "5e7da9b618d69acebed284c2711d026acc1ca551"
+      ref: gradient_avatar
```
wal33d006, line 71: @andrzejchm 
Added this because I added a function in `cosmos_utils`. Should have been a small separate PR but since we are already into nested `starport-redesign` Pas so I thought I should update the reference, will update it back to `starport-redesign` in the next PR

andrzejchm (approved):

PR #134, opened 2021-11-27
feat: Back up wallet page - components

andrzejchm:


packages/cosmos_ui_components/lib/components/mnemonic_choice_chip.dart
```diff
@@ -17,42 +18,54 @@ class MnemonicChoiceChip extends StatelessWidget {
...
         ),
       ),
     );
   }
+
+  Widget _indexText(BuildContext context, int index) => Padding(
+        padding: EdgeInsets.only(right: CosmosTheme.of(context).spacingM),
+        child: Container(
+          width: 18,
+          height: 18,
```
wal33d006, line 56: Do we have a special requirement of an 18x18 text box?

wal33d006 (approved):
Just one comment, other than that good to go!

packages/cosmos_ui_components/lib/components/mnemonic_choice_chip.dart
andrzejchm: its according to the figma design:
image

wal33d006: Perfect

PR #136, opened 2021-11-29
fix: Bottom sheet header fix

wal33d006:


packages/cosmos_ui_components/lib/components/cosmos_bottom_sheet_header.dart
```diff
@@ -23,6 +23,7 @@ class CosmosBottomSheetHeader extends StatelessWidget {
     return Column(
       mainAxisSize: MainAxisSize.min,
       children: [
+        SizedBox(height: CosmosTheme.of(context).spacingL),
```
andrzejchm, line 26: while this adds spacing to the header for situations with no `leading` nor `actions`, it breaks the designs for those with leading/actions:
image
CleanShot 2021-11-29 at 13 19 56@2x
Also, the spacing is too small according to designs:
CleanShot 2021-11-29 at 13 21 05@2x

andrzejchm (changes requested):

packages/cosmos_ui_components/lib/components/cosmos_bottom_sheet_header.dart
wal33d006: It doesn't break the design for those with leading/actions. I will remove that extra spacing from the template in another short PR. Also, 40 isn't something which we have in spacings. It is either 16 or 32. Should I create one?

andrzejchm: ah, yes, it looks perfect on the sheet with leading/actions. Could you create the `spacingXXL` for 40 px then? :)

wal33d006: Thanks, updated :)

andrzejchm (approved):

PR #138, opened 2021-11-30
feat: Add function for updating public wallet info using plain data store

wal33d006:


packages/transaction_signing_gateway/lib/gateway/transaction_signing_gateway.dart
```diff
@@ -57,6 +57,12 @@ class TransactionSigningGateway {
         password: password,
       );
 
+  /// Updates the public details of the wallet using [PlainDataStore]
```
andrzejchm, line 60: no need to add the info regarding `PlainDataStore`. this is an implementation detail and its better to not give a hint about that to the library users

packages/transaction_signing_gateway/lib/storage/cosmos_key_info_storage.dart
```diff
@@ -167,4 +167,28 @@ class CosmosKeyInfoStorage implements KeyInfoStorage {
...
+
+      return _plainDataStore.readPlainText(key: publicInfoKey).flatMap((walletInfo) async {
+        if (walletInfo == null) {
+          return left(const CredentialsStorageFailure("Wallet not found"));
+        }
+
+        final publicInfoJson = await compute(jsonEncode, WalletPublicInfoSerializer.toMap(info));
+        await _plainDataStore.savePlainText(key: publicInfoKey, value: publicInfoJson);
+
+        return right(unit);
```
andrzejchm, line 187: you're ignoring potential errors in `savePlainText` and always returning `right(unit)`. you should return result of `savePlainText` directly so that the resulting future will fail with `left(..)` if saving fails
```suggestion
        return _plainDataStore.savePlainText(key: publicInfoKey, value: publicInfoJson);
```

andrzejchm (approved):
two minor comments and we're good to go!

packages/transaction_signing_gateway/lib/storage/cosmos_key_info_storage.dart
wal33d006: Got it.

packages/transaction_signing_gateway/lib/gateway/transaction_signing_gateway.dart
wal33d006: Perfect. Updated this

andrzejchm (approved):

PR #139, opened 2021-11-30
feat: Repeat mnemonic page - components

andrzejchm:

### DIFF
--- a/starport_template/lib/pages/routing_page.dart
+++ b/starport_template/lib/pages/routing_page.dart
@@ -25,7 +25,7 @@ class _RoutingPageState extends State<RoutingPage> {
       if (!mounted) {
         return;
       }
-      if (store.wallets.isNotEmpty) {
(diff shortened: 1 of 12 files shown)


starport_template/lib/pages/routing_page.dart
```diff
@@ -25,7 +25,7 @@ class _RoutingPageState extends State<RoutingPage> {
       if (!mounted) {
         return;
       }
-      if (store.wallets.isEmpty) {
+      if (store.wallets.isNotEmpty) {
```
wal33d006, line 28: Why did we update this condition?

wal33d006 (approved):
Just one comment.

starport_template/lib/pages/routing_page.dart
andrzejchm: good catch! I always forget about changing this back

PR #141, opened 2021-11-30
feat: Rename wallet - Components

wal33d006:


packages/cosmos_ui_components/lib/components/cosmos_bottom_sheet_header.dart
```diff
@@ -19,14 +19,14 @@ class CosmosBottomSheetHeader extends StatelessWidget {
...
   @override
   Widget build(BuildContext context) {
+    final isExtended = leading != null || actions != null;
     final theme = CosmosTheme.of(context);
     return Column(
       mainAxisSize: MainAxisSize.min,
       children: [
-        SizedBox(height: CosmosTheme.of(context).spacingXXL),
-        if (leading != null || actions != null)
+        if (isExtended)
```
wal33d006, line 27: The bottom sheet header had issues. You were right about extra spacing when we have leading/actions. I fixed that properly now with all the combinations.

packages/cosmos_ui_components/lib/components/cosmos_bottom_sheet_header.dart
```diff
@@ -42,7 +42,10 @@ class CosmosBottomSheetHeader extends StatelessWidget {
               ],
             ),
           ),
-        SizedBox(height: theme.spacingM),
+        if (isExtended)
+          Column(children: [SizedBox(height: theme.spacingM), SizedBox(height: theme.spacingL)])
```
andrzejchm, line 46: won't this work ?:
```suggestion
          SizedBox(height: theme.spacingM + theme.spacingL)
```

Also, if you need some sppacing that is between L and XL, you can simply rename them and include one additional, i.e:
rename:
XL -> XXL
XXL -> XXXL
add new:
XL
We'll deal with name discrepancies in Emeris later on

andrzejchm: aaand to optimize the code even more, you can do this:
```dart
SizedBox(height: isExteneded ? theme.spacingXL : theme.spacingXXXL)
```

packages/cosmos_ui_components/lib/components/cosmos_circle_text_button.dart
```diff
@@ -5,32 +5,40 @@ class CosmosCircleTextButton extends StatelessWidget {
   final String text;
   final VoidCallback onTap;
   final IconData icon;
+  final bool isEnabled;
```
andrzejchm, line 8: to control enabled/disabled state lets follow flutter's approach, where setting `onTap` to null makes the button disabled

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
```diff
@@ -0,0 +1,60 @@
...
+    this.text = '',
+    this.maxLines,
+  }) : super(key: key);
+
+  @override
+  State<CosmosTextField> createState() => _CosmosTextFieldState();
+}
+
+class _CosmosTextFieldState extends State<CosmosTextField> {
+  final TextEditingController controller = TextEditingController();
```
andrzejchm, line 23: you should initialize controller in `initState` and dispose it in `dispose`

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
```diff
@@ -0,0 +1,60 @@
...
+          width: 17,
+          child: CircleAvatar(
+            backgroundColor: theme.colors.divider,
+            foregroundColor: theme.colors.background,
+            child: InkWell(
+              onTap: () {
+                controller.clear();
+                widget.onChanged('');
+              },
+              child: const Icon(Icons.clear, size: 12),
```
andrzejchm, line 53: we should refrain from using default flutter icons, as those are subject to change and differ between platforms. We should always pull the icons from designs and place them in assets, take a look into `CosmosBackButton` and `"assets/images/big-arrow-left.png"` for example

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
```diff
@@ -0,0 +1,60 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final theme = CosmosTheme.of(context);
+    return TextField(
+      controller: controller,
+      maxLines: widget.maxLines,
+      maxLength: widget.maxLength,
+      onChanged: widget.onChanged,
+      decoration: InputDecoration(
+        border: const UnderlineInputBorder(),
```
andrzejchm, line 40: are we sure the default colors match the ones specified in figma?

packages/cosmos_ui_components/lib/components/cosmos_wallet_edit_button.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:cosmos_ui_components/cosmos_ui_components.dart';
+import 'package:flutter/material.dart';
+
+class CosmosWalletEditButton extends StatelessWidget {
```
andrzejchm, line 4: this is a bit too specific component to be used in `cosmos_ui_components`. Also, this could be simply exported from figma as an asset and used inside `Image.asset` wrapped with `InkWell`, something simmilar to what `IconButton` in flutter does. We could create `CosmosImageButton` actually that allows you to pass in the path to the asset you want to use as the button

packages/cosmos_ui_components/lib/components/template/cosmos_wallet_list_item.dart
```diff
@@ -1,53 +1,67 @@
...
 import 'package:cosmos_ui_components/cosmos_theme.dart';
 import 'package:flutter/material.dart';
 
 class CosmosWalletListItem extends StatelessWidget {
   final String name;
   final String address;
   final bool isSelected;
+  final bool isEditing;
   final VoidCallback onClicked;
+  final Function(String) onEditIconPressed;
```
andrzejchm, line 13: to make it more generic we should just allow to pass in the `suffix`. if one needs custom logic, it can be implemented outside of the component itself. Also, the `Icons.check_circle_outline` should be replaced with an asset from Figma. and it should not be visible by default, but only for currently selected wallet

packages/cosmos_ui_components/lib/cosmos_theme.dart
```diff
@@ -161,6 +162,7 @@ class CosmosColorsData extends Equatable {
   final Color chipBackground;
   final Color positiveText;
   final Color link;
+  final Color clearButtonBg;
```
andrzejchm, line 165: this color name is again too specific, we won't be able to include all the colors inside the designs into theme and should instead "generalize" the pallete as much as possible, also I can't see this color used anywhere

packages/cosmos_ui_components/lib/cosmos_text_theme.dart
```diff
@@ -63,5 +63,11 @@ abstract class CosmosTextTheme {
         height: 1.23,
       );
 
+  static TextStyle get labelM => const TextStyle(
```
andrzejchm, line 66: we take the typography namings straight from figma, as far as I can see, the wallet name uses `title0Medium`
CleanShot 2021-11-30 at 17 51 52@2x

packages/cosmos_ui_components/lib/components/template/cosmos_wallet_list_item.dart
```diff
@@ -1,53 +1,67 @@
...
 import 'package:cosmos_ui_components/components/gradient_avatar.dart';
+import 'package:cosmos_ui_components/cosmos_text_theme.dart';
 import 'package:cosmos_ui_components/cosmos_theme.dart';
 import 'package:flutter/material.dart';
 
 class CosmosWalletListItem extends StatelessWidget {
   final String name;
   final String address;
   final bool isSelected;
+  final bool isEditing;
```
andrzejchm, line 11: whe adding new parameters to components, lets refrain from marking them as `required` but instead provide the default value to make it as much backward-compatibile as possible

andrzejchm (changes requested):
let me know if you have any questions regarding the comments, we can trough them over slack

packages/cosmos_ui_components/lib/cosmos_text_theme.dart
andrzejchm: Ah, that makes sense. In one of the places in figma I found some `label/X` styles to be marked as deprecated, so lets use title0Medium then if possible

packages/cosmos_ui_components/lib/components/cosmos_bottom_sheet_header.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_circle_text_button.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
wal33d006: Yeah my bad, missed that color. Updated

packages/cosmos_ui_components/lib/components/cosmos_wallet_edit_button.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/template/cosmos_wallet_list_item.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/cosmos_theme.dart
wal33d006: Removed.

packages/cosmos_ui_components/lib/components/template/cosmos_wallet_list_item.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_bottom_sheet_header.dart
```diff
@@ -42,7 +42,7 @@ class CosmosBottomSheetHeader extends StatelessWidget {
               ],
             ),
           ),
-        SizedBox(height: theme.spacingM),
+        SizedBox(height: isExtended ? theme.spacingXL : theme.spacingXXXL),
         Padding(
```
andrzejchm, line 46: since we already have padding widget below, we could simply add `top: isExtended ? theme.spacingXL : theme.spacingXXXL` and remove the sizedBox alltogether

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:cosmos_ui_components/cosmos_ui_components.dart';
+import 'package:flutter/material.dart';
+
+class CosmosImageButton extends StatelessWidget {
+  final String assetUrl;
```
andrzejchm, line 5: its rather a path than url, but you can just simply name it `asset`, like `Image.asset()` does

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
```diff
@@ -0,0 +1,31 @@
...
+    required this.onPressed,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return InkWell(
+      onTap: () => onPressed(),
+      child: SizedBox(
+        height: 24,
+        width: 24,
```
andrzejchm, line 20: we could parametrize the width and height as I'm pretty sure we'll have many different sizes for the button

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
```diff
@@ -0,0 +1,31 @@
...
+        child: Container(
```
andrzejchm, line 21: I'm not sure if order should be part of the ImageButton

packages/cosmos_ui_components/lib/components/template/cosmos_wallet_list_item.dart
```diff
@@ -7,47 +8,54 @@ class CosmosWalletListItem extends StatelessWidget {
...
+          Container(
+            padding: EdgeInsets.all(theme.spacingL),
+            margin: EdgeInsets.symmetric(
+              horizontal: theme.spacingL,
+              vertical: theme.spacingM,
+            ),
+            decoration: BoxDecoration(
+              borderRadius: theme.borderRadiusM,
+              // TODO: This color needs to be fixed. Doesn't match the dark theme.
+              border: Border.all(color: isSelected ? Colors.black12 : Colors.transparent),
```
andrzejchm, line 38: we should use theme here for the colors. this way we will automatically start supporting dark theme when we introduce separate theme values for dark mode

packages/cosmos_ui_components/lib/components/template/cosmos_wallets_list_view.dart
```diff
@@ -6,27 +7,38 @@ class CosmosWalletsListView extends StatelessWidget {
...
+            .mapIndexed(
+              (index, wallet) => CosmosWalletListItem(
+                name: wallet.name,
+                address: wallet.address,
+                isSelected: selectedWallet.address == wallet.address,
+                onClicked: () => onClicked(index),
+                suffix: isEditing
+                    ? CosmosImageButton(assetUrl: 'assets/images/more.png', onPressed: () => onEditIconPressed(wallet))
+                    : selectedWallet.address == wallet.address
+                        ? CosmosImageButton(assetUrl: 'assets/images/check.png', onPressed: () {})
```
andrzejchm, line 36: this is not a button but rather an indicator, so you should either use plain `Image.asset` or at least pass `null` to `onPressed`, otherwise there will be splash effect visible when touching

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
```diff
@@ -0,0 +1,31 @@
...
+import 'package:flutter/material.dart';
+
+class CosmosImageButton extends StatelessWidget {
+  final String assetUrl;
+  final VoidCallback onPressed;
+
+  const CosmosImageButton({
+    Key? key,
+    required this.assetUrl,
+    required this.onPressed,
```
andrzejchm, line 11: could we follow other widgets here and use `onTapXX` naming? here, simply `onTap` will be perfect.
(other widgets can have more specialized naming, like `onTapWallet`, `onTapEdit` etc. rather than `onPressedWallet` `onEditPressed` etc.)

andrzejchm (changes requested):

packages/cosmos_ui_components/lib/components/cosmos_bottom_sheet_header.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/template/cosmos_wallet_list_item.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
```diff
@@ -1,30 +1,27 @@
...
+  final double height;
+  final double width;
+  final VoidCallback? onTap;
 
   const CosmosImageButton({
     Key? key,
-    required this.assetUrl,
-    required this.onPressed,
+    required this.asset,
+    required this.onTap,
```
andrzejchm, line 12: its nullable field, we can remove the `required` keyword here

wal33d006: Updated

andrzejchm (approved):
great work! 🚀

PR #143, opened 2021-12-01
fix: Update package name in assets

wal33d006:


packages/cosmos_ui_components/lib/components/cosmos_circle_text_button.dart
```diff
@@ -26,13 +28,17 @@ class CosmosCircleTextButton extends StatelessWidget {
...
-              maxRadius: 12,
-              minRadius: 12,
-              backgroundColor: isEnabled ? theme.colors.text : theme.colors.inactive,
-              foregroundColor: Colors.white,
-              child: Icon(icon, size: theme.fontSizeM),
+            SizedBox(
+              height: 24,
+              width: 24,
+              child: ClipRRect(
+                borderRadius: BorderRadius.circular(15),
```
wal33d006, line 35: Wrapped the image around this because when giving color to image, it was appearing in a square shape. Do let me know if there is any better way of doing this

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
```diff
@@ -21,7 +22,7 @@ class CosmosImageButton extends StatelessWidget {
       child: SizedBox(
         height: height,
         width: width,
-        child: Image.asset(asset),
+        child: Image.asset(asset, package: packageName),
```
andrzejchm, line 25: this won't work in this particular example. if someone tries to use the `CosmosImageButton` with an asset outside of the `cosmos_ui_components`. what you should do here is to allow specify the package as a constructor param that defaults to null and provide the packageName at the call site

packages/cosmos_ui_components/lib/components/cosmos_circle_text_button.dart
```diff
@@ -3,14 +3,16 @@ import 'package:flutter/material.dart';
 
 class CosmosCircleTextButton extends StatelessWidget {
   final String text;
+  final String asset;
   final VoidCallback? onTap;
   final IconData icon;
```
andrzejchm, line 8: you're not using `icon` anymore, lets remove it

packages/cosmos_ui_components/lib/components/cosmos_circle_text_button.dart
wal33d006: Setting color to asset image will paint all pixels right. But using `colorBlendMode` will fix that. But that will result in putting that color blend to the whole container (which is a square). to prevent that I rounded the container around it.

packages/cosmos_ui_components/lib/components/cosmos_circle_text_button.dart
wal33d006: Removed

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
wal33d006: Updated

andrzejchm: you now also have to update call sites (i.e: CosmosWalletsListView) to pass in the proper `package`

packages/cosmos_ui_components/lib/components/cosmos_circle_text_button.dart
wal33d006: Fixed this now with latest PNGs being sent from the consumer end and applying simple color

packages/cosmos_ui_components/lib/components/cosmos_image_button.dart
wal33d006: Updated

andrzejchm (approved):
🚀 🚢

PR #144, opened 2021-12-02
feat: Transaction signing - Components

wal33d006:


packages/cosmos_ui_components/lib/components/cosmos_divider.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:cosmos_ui_components/cosmos_ui_components.dart';
+import 'package:flutter/material.dart';
+
+class CosmosDivider extends StatelessWidget {
+  const CosmosDivider({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = CosmosTheme.of(context);
+    return Padding(
```
andrzejchm, line 10: I'm pretty sure this padding will differ on different screen, so what I'd do is to pass in the `EdgeInsets` as param with default value of `null`:

```dart
...
final EdgeInsets? edgeInsets;

const CosmosDivider({
    Key? key, 
   this.edgeInsets,
}) : super(key: key);

... 
    ...
    padding: edgeInsets ?? EdgeInsets.symmetric(horizontal: theme.spacingL)
    ...
}

```

packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
```diff
@@ -30,12 +32,16 @@ class CosmosElevatedButton extends StatelessWidget {
...
+            prefixIcon!,
+            SizedBox(width: CosmosTheme.of(context).spacingM),
+          ],
           Text(
             text,
             style: CosmosTextTheme.elevatedButton,
           ),
           if (suffixIcon != null) ...[
-            SizedBox(width: CosmosTheme.of(context).spacingS),
+            SizedBox(width: CosmosTheme.of(context).spacingM),
```
andrzejchm, line 44: this changes current design for other places, if you need other spacing in a particular case then lets parametrize it with the default value of `spacingS`

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
```diff
@@ -42,17 +49,33 @@ class _CosmosTextFieldState extends State<CosmosTextField> {
       controller: controller,
       maxLines: widget.maxLines,
       maxLength: widget.maxLength,
-      onChanged: widget.onChanged,
+      keyboardType: widget.keyboardType,
+      onChanged: (value) {
+        widget.onChanged(value);
+        setState(() {});
```
andrzejchm, line 55: this will cause a lot of rebuilds on each character change. could you optimize the logic and call the `setState` only if the text changes from empty to nonempty state and vice versa?

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
```diff
@@ -42,17 +49,33 @@ class _CosmosTextFieldState extends State<CosmosTextField> {
...
         counterText: '',
-        suffix: InkWell(
-          onTap: () {
-            controller.clear();
-            widget.onChanged('');
-          },
-          child: SizedBox(height: 17, width: 17, child: Image.asset('assets/images/cross.png', package: packageName)),
-        ),
+        suffixIcon: controller.text.isNotEmpty ? null : widget.suffix,
+        suffix: widget.suffix != null ? (controller.text.isNotEmpty ? _buildClearButton() : null) : _buildClearButton(),
```
andrzejchm, line 63: the general rule of if/else statements is that the condition inside if statement should be non-negated in order to avoid double negation, i.e:

instead of writing
```
if(false) {
  "its false"
} else {
"its not false" => "its true'
}
```

do this:

```
if(true) {
   "true!"
 } else {
   "false"
}
```

this is because double negation is harder to comprehend than just a true statement

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
```diff
@@ -42,17 +49,33 @@ class _CosmosTextFieldState extends State<CosmosTextField> {
...
+        hintStyle: CosmosTextTheme.copy0Normal,
         counterText: '',
-        suffix: InkWell(
-          onTap: () {
-            controller.clear();
-            widget.onChanged('');
-          },
-          child: SizedBox(height: 17, width: 17, child: Image.asset('assets/images/cross.png', package: packageName)),
-        ),
+        suffixIcon: controller.text.isNotEmpty ? null : widget.suffix,
```
andrzejchm, line 62: same here with double negation (see below)

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
```diff
@@ -42,17 +49,33 @@ class _CosmosTextFieldState extends State<CosmosTextField> {
...
+        suffix: widget.suffix != null ? (controller.text.isNotEmpty ? _buildClearButton() : null) : _buildClearButton(),
+      ),
+    );
+  }
+
+  InkWell _buildClearButton() {
+    return InkWell(
+      onTap: () {
+        controller.clear();
+        widget.onChanged('');
```
andrzejchm, line 72: you already call `onChanged` on each character change, so no need to call it again here explicitly

packages/cosmos_ui_components/lib/cosmos_text_theme.dart
```diff
@@ -17,6 +17,19 @@ abstract class CosmosTextTheme {
         height: 1.27,
       );
 
+  TextStyle get titleS => const TextStyle(
+        fontFamily: "Inter",
+        fontWeight: FontWeight.w700,
+        fontSize: 13,
+        height: 1.27,
```
andrzejchm, line 24: as far as I can see in the designs, the line height is `15.99` which divided by font size gives: `15.99/13 = 1.23`

packages/cosmos_ui_components/lib/cosmos_text_theme.dart
```diff
@@ -17,6 +17,19 @@ abstract class CosmosTextTheme {
...
+      );
+
+  TextStyle get title1Medium => const TextStyle(
+        fontFamily: "Inter",
+        fontWeight: FontWeight.w600,
+        fontSize: 21,
```
andrzejchm, line 30: this should have `height = 1.287`

packages/cosmos_ui_components/lib/components/cosmos_token_avatar.dart
```diff
@@ -0,0 +1,19 @@
...
+  final String text;
+
+  const CosmosTokenAvatar({Key? key, required this.text}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return SizedBox(
+      height: 32,
+      child: CircleAvatar(
+        backgroundColor: Colors.grey[300],
```
andrzejchm, line 13: we should always use colors from the theme, otherwise we will lack the dark mode/ light mode support and have to pinpoint all those places in code to update it

packages/cosmos_ui_components/lib/components/cosmos_divider.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_text_field.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/cosmos_text_theme.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/cosmos_token_avatar.dart
wal33d006: Updated

andrzejchm (approved):

PR #146, opened 2021-12-03
fix: Update cosmos_balance_card.dart

wal33d006:


packages/cosmos_ui_components/lib/components/template/cosmos_balance_card.dart
```diff
@@ -1,62 +1,62 @@
...
                 ),
+              ],
+            ),
+            if (!isListTileType)
+              Column(
+                crossAxisAlignment: CrossAxisAlignment.end,
+                mainAxisAlignment: MainAxisAlignment.center,
+                children: [
+                  Text(amountDisplayText, style: CosmosTextTheme.labelS),
+                  Text('available ${denomText.toUpperCase()}', style: CosmosTextTheme.copyMinus1Normal),
```
andrzejchm, line 55: this hardcodes the text and does not allow for locallization. we should allow to just pass `secondaryText` here which is formatted on the call site

packages/cosmos_ui_components/lib/components/template/cosmos_balance_card.dart
```diff
@@ -1,62 +1,62 @@
...
+                  ],
                 ),
+              ],
+            ),
+            if (!isListTileType)
+              Column(
+                crossAxisAlignment: CrossAxisAlignment.end,
+                mainAxisAlignment: MainAxisAlignment.center,
+                children: [
+                  Text(amountDisplayText, style: CosmosTextTheme.labelS),
```
andrzejchm, line 54: are we sure we want to have `amountDisplayText` shown twice in the same item? here and on line 39

wal33d006: This is a different condition. Actually there are two styles in Figma for card for the same type of data. There is an `if` condition at line 49.
Screenshot 2021-12-03 at 1 58 13 PM

packages/cosmos_ui_components/lib/components/template/cosmos_balance_card.dart
wal33d006: Updated

packages/cosmos_ui_components/lib/components/template/cosmos_balance_card.dart
andrzejchm: ah, makes sense! Didn't notice the first `if` statement. my bad! 🙏

packages/cosmos_ui_components/lib/components/template/cosmos_balance_card.dart
```diff
@@ -52,7 +54,7 @@ class CosmosBalanceCard extends StatelessWidget {
                 mainAxisAlignment: MainAxisAlignment.center,
                 children: [
                   Text(amountDisplayText, style: CosmosTextTheme.labelS),
-                  Text('available ${denomText.toUpperCase()}', style: CosmosTextTheme.copyMinus1Normal),
+                  Text('$secondaryText ${denomText.toUpperCase()}', style: CosmosTextTheme.copyMinus1Normal),
```
andrzejchm, line 57: what I meant here was to make the entire text a `secondaryText`, since in some languages the word order might be different.

```suggestion
                  Text(secondaryText, style: CosmosTextTheme.copyMinus1Normal),
```

wal33d006: Updated. I made the whole `secondaryText` a required parameter.

andrzejchm: but it won't make much sense to be required, if you fall into case where its not displayed at all, maybe just make it default to `""`?

wal33d006: Yeah you are right, my bad this should be optional. Just updated it

andrzejchm (approved):

PR #154, opened 2021-12-09
feat: add TextEditingController to CosmosTextField 

andrzejchm:
This should solve our problems with paste handling and user typing


wal33d006 (approved):
Awesome!

PR #158, opened 2021-12-13
feat: Transaction history - Components

wal33d006:
Screenshot 2021-12-13 at 11 43 55 PM


packages/cosmos_ui_components/lib/utils/group_by_extension.dart
```diff
@@ -0,0 +1,6 @@
+extension Iterables<E> on Iterable<E> {
+  Map<K, List<E>> groupBy<K>(K Function(E) keyFunction) => fold(
+        <K, List<E>>{},
+        (Map<K, List<E>> map, E element) => map..putIfAbsent(keyFunction(element), () => <E>[]).add(element),
+      );
```
andrzejchm, line 5: cool addition, could we actually put it in `cosmos_utils` so its available for broader use?

packages/cosmos_ui_components/lib/utils/date_formatter.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:intl/intl.dart';
+
+String formatDate(DateTime date, DateFormatEnum dateFormat) => DateFormat(getDateFormat[dateFormat]).format(date);
+
+abstract class DateFormats {
+  static const String UIShortDate = 'MMM d';
```
andrzejchm, line 6: variables should start with lowercase letters:
```suggestion
  static const String uiShortDate = 'MMM d';
```

packages/cosmos_ui_components/lib/components/template/transaction_history_group_title.dart
```diff
@@ -0,0 +1,20 @@
...
+  const TransactionHistoryGroupTitle({
+    Key? key,
+    required this.title,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Padding(
+      padding: EdgeInsets.all(CosmosTheme.of(context).spacingL),
+      child: Text(title.split('-')[0], style: CosmosTextTheme.title0Bold),
```
andrzejchm, line 17: why do we need this title splitting here? what will happen if we pass a title that has no `-` inside?

packages/cosmos_ui_components/lib/components/template/transaction_history_card.dart
```diff
@@ -0,0 +1,71 @@
...
+                  ),
+                ],
+              ),
+              const Spacer(),
+              Column(
+                crossAxisAlignment: CrossAxisAlignment.end,
+                children: [
+                  Text(
+                    '${isOutgoing ? '-' : '+'} $amountText',
+                    style: CosmosTextTheme.labelS.copyWith(color: isOutgoing ? null : Colors.green),
```
andrzejchm, line 56: hardcoded color

packages/cosmos_ui_components/lib/utils/date_formatter.dart
```diff
@@ -0,0 +1,15 @@
...
+  static const String groupByMonthYear = 'MMMM-yyyy';
+}
+
+enum DateFormatEnum { ShortUIDateDay, GroupByMonthYear }
```
andrzejchm, line 10: lets add trailing commas wherever possible for readability.
Also, most of the enum cases in our code start with lowercase letter, lets stick with that

packages/cosmos_ui_components/lib/components/template/transaction_history_card.dart
```diff
@@ -0,0 +1,71 @@
...
+  Widget build(BuildContext context) {
+    final theme = CosmosTheme.of(context);
+    return Padding(
+      padding: EdgeInsets.symmetric(horizontal: theme.spacingL, vertical: theme.spacingM),
+      child: Column(
+        children: [
+          Row(
+            children: [
+              Image.asset(
+                isOutgoing ? 'assets/images/send_icon.png' : 'assets/images/receive_icon.png',
```
andrzejchm, line 32: we can optimize it and instead of having two separate assets for the same icon, we can simply add rotation to the image based on the `isOutgoing` bool value

andrzejchm (changes requested):
nice work! I've added some comments with improvements

packages/cosmos_ui_components/lib/components/template/transaction_history_group_title.dart
wal33d006: Yes, my bad. This split is because I am grouping the list by month+year with a '-' in between. This needs to go out to the consumer end. Will remove this from here.

packages/cosmos_ui_components/lib/components/template/transaction_history_card.dart
wal33d006: The color for incoming transactions will always be green. Should I create a green color inside the theme?

andrzejchm: yes, we should always grab the colors from theme. especially that this should be `#008223` rather than `Colors.green`. Also, we can't assume that all implementations of the app will use green color for incomming transactions. some might theme it as blue for example, or use different shade of green for dark theme

packages/cosmos_ui_components/lib/cosmos_theme.dart
```diff
@@ -167,6 +168,7 @@ class CosmosColorsData extends Equatable {
   final Color actionSheetDestructive;
   final Color chipBackground;
   final Color positiveText;
+  final Color brightText;
```
andrzejchm, line 171: we already have `positiveText` that has this color, no need to duplicate it

andrzejchm (changes requested):
few more updates

andrzejchm (approved):

PR #159, opened 2021-12-13
feat: Transaction history

wal33d006:
Screenshot 2021-12-13 at 11 43 55 PM


starport_template/lib/widgets/back_up_account_card.dart
```diff
@@ -0,0 +1,36 @@
...
+      child: Container(
+        margin: EdgeInsets.all(theme.spacingL),
+        padding: EdgeInsets.symmetric(horizontal: theme.spacingL, vertical: theme.spacingM),
+        decoration: BoxDecoration(
+          borderRadius: BorderRadius.circular(10),
+          border: Border.all(color: theme.colors.inactive),
+        ),
+        child: Row(
+          children: [
+            Image.asset('assets/images/warning.png'),
```
andrzejchm, line 26: we already have that icon in `cosmos_ui_components` as `icon_warning.png`. we also have such component called `CosmosWarningBox` that uses it. you can adapt it to also include suffix arrow optionally

starport_template/lib/entities/transaction.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:starport_template/entities/amount.dart';
+import 'package:starport_template/entities/denom.dart';
+
+enum TransactionType {
+  Receive,
+  Send,
+}
+
+class Transaction {
```
andrzejchm, line 9: how much different is it from the `MsgSendTransaction`? could we somehow reuse it here? or choose a different naming that explains the difference between `Transaction` and `MsgSendTransaction`

starport_template/lib/model/tx_json.dart
```diff
@@ -0,0 +1,59 @@
+class Tx {
```
andrzejchm, line 1:
```suggestion
class TxJson {
```

starport_template/lib/model/tx_json.dart
```diff
@@ -0,0 +1,59 @@
...
+  String type;
+  Body body;
+
+  factory Tx.fromJson(Map<String, dynamic> json) => Tx(
+        type: json["@type"] as String? ?? '',
+        body: Body.fromJson(json["body"] as Map<String, dynamic>),
+      );
+}
+
+class Body {
```
andrzejchm, line 16:
```suggestion
class TxBodyJson {
```
also, should be in separate file

starport_template/lib/model/tx_json.dart
```diff
@@ -0,0 +1,59 @@
...
+  Body({required this.messages});
+
+  List<Message> messages;
+
+  factory Body.fromJson(Map<String, dynamic> json) =>
+      Body(messages: (json["messages"] as List).map((x) => Message.fromJson(x as Map<String, dynamic>)).toList());
+}
+
+class Message {
```
andrzejchm, line 25:
```suggestion
class TxBodyMessageJson {
```
also, should be in separate file

starport_template/lib/model/tx_json.dart
```diff
@@ -0,0 +1,59 @@
...
+
+  factory Message.fromJson(Map<String, dynamic> json) => Message(
+        type: json["@type"] as String? ?? '',
+        fromAddress: json["from_address"] as String? ?? '',
+        toAddress: json["to_address"] as String? ?? '',
+        amount: (json["amount"] as List).map((x) => Amount.fromJson(x as Map<String, dynamic>)).toList(),
+      );
+}
+
+class Amount {
```
andrzejchm, line 46:
```suggestion
class AmountJson {
```
in separate file

starport_template/lib/model/tx_response_json.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:starport_template/model/tx_json.dart';
+
+class TxResponse {
```
andrzejchm, line 3:
```suggestion
class TxResponseJson {
```

starport_template/lib/pages/assets_portfolio_page.dart
```diff
@@ -79,16 +80,27 @@ class _AssetsPortfolioPageState extends State<AssetsPortfolioPage> {
...
+            child: GradientAvatar(stringKey: selectedWallet.publicAddress),
+          ),
+        ),
       ),
     );
   }
 
+  Future<void> _onTapAvatar(BuildContext context) async {
+    await Navigator.of(context).push(MaterialPageRoute(builder: (context) => const TransactionHistoryPage()));
+    StarportApp.walletsStore.getBalances(selectedWallet.publicAddress);
```
andrzejchm, line 101: do we need to fetch balances after the `TransactionHistoryPage` is closed?

starport_template/lib/stores/wallets_store.dart
```diff
@@ -45,6 +50,14 @@ class WalletsStore {
...
   set areWalletsLoading(bool val) => Action(() => _areWalletsLoading.value = val)();
 
+  bool get isTransactionHistoryLoading => _isTransactionHistoryLoading.value;
+
+  set isTransactionHistoryLoading(bool val) => Action(() => _isTransactionHistoryLoading.value = val)();
+
+  bool get isTransactionHistoryError => _isTransactionHistoryError.value;
+
+  set isTransactionHistoryError(bool val) => Action(() => _isTransactionHistoryError.value = val)();
+
```
andrzejchm, line 60: the `WalletsStore` is getting really bloated with code. can we extract this code to separate `TransactionsStore` instead?

starport_template/lib/utils/cosmos_transaction_history.dart
```diff
@@ -0,0 +1,47 @@
+import 'dart:convert';
+
+import 'package:http/http.dart' as http;
+import 'package:starport_template/entities/amount.dart';
+import 'package:starport_template/entities/denom.dart';
+import 'package:starport_template/entities/transaction.dart';
+import 'package:starport_template/model/tx_response_json.dart';
+import 'package:starport_template/utils/base_env.dart';
+
+class CosmosTransactionHistory {
```
andrzejchm, line 10: this is not an entity holding the history of transactions, its more of a repository or `HistoryLoader`

starport_template/lib/utils/cosmos_transaction_history.dart
```diff
@@ -0,0 +1,47 @@
...
+    final outGoingTransactions = await getTransactionResponses(walletAddress, TransactionType.Send);
+    final incomingTransactions = await getTransactionResponses(walletAddress, TransactionType.Receive);
+
+    final list = [...outGoingTransactions, ...incomingTransactions];
+    list.sort((a, b) => b.date.compareTo(a.date));
+
+    return list;
+  }
+
+  Future<List<Transaction>> getTransactionResponses(String walletAddress, TransactionType type) async {
```
andrzejchm, line 25: this should be private

starport_template/lib/widgets/asset_portfolio_heading.dart
```diff
@@ -4,14 +4,17 @@ import 'package:flutter/material.dart';
 class AssetPortfolioHeading extends StatelessWidget {
   final String title;
   final VoidCallback onTap;
+  final bool isCentered;
 
-  const AssetPortfolioHeading({Key? key, required this.title, required this.onTap}) : super(key: key);
+  const AssetPortfolioHeading({Key? key, required this.title, required this.onTap, this.isCentered = false})
```
andrzejchm, line 9: trailing comma, its hard to read such inlined constructor

starport_template/lib/widgets/transaction_history_list.dart
```diff
@@ -0,0 +1,55 @@
...
+          final transactionsList = groupedHistory[groupedMonthYear] ?? [];
+          return SliverToBoxAdapter(
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+                TransactionHistoryGroupTitle(title: groupedMonthYear.split('-')[0]),
+                ...transactionsList
+                    .map(
+                      (transaction) => TransactionHistoryCard(
+                        transactionType: transaction.type.toString().split('.')[1],
```
andrzejchm, line 40: splitting and parsing strings should not be part of UI, this should be a getter inside the `Transaction`type

starport_template/lib/stores/wallets_store.dart
```diff
@@ -268,4 +281,18 @@ class WalletsStore {
       logError(ex, stack);
     }
   }
+
+  Future<void> getTransactionHistory() async {
+    transactionsList.clear();
+    isTransactionHistoryLoading = true;
+    try {
+      transactionsList.addAll(
```
andrzejchm, line 289: if it fails, you'll end up with cleared list and no transactions. you should clear the list just before adding new elements to it

andrzejchm (changes requested):

starport_template/lib/entities/transaction.dart
wal33d006: The only common thing between the two is the `amount` parameter. The `Transaction` is something which is populated for transactions history so it's quite different from `MsgSendTransaction`

wal33d006: We can name it `TransactionHistory` maybe

andrzejchm: `TransactionHistoryItem`?

starport_template/lib/pages/assets_portfolio_page.dart
wal33d006: The `selectedWallet` might be changed by the user in this history page and it needs to reflect the latest balances from the current selected wallet

andrzejchm: we should make it a part of changing the selected wallet then, so that each change will trigger this reload, instead of doing it in UI. you can make it inside the setter for `selectedWalletIndex` probably

starport_template/lib/pages/transaction_history_page.dart
```diff
@@ -0,0 +1,159 @@
...
+          children: [
+            _appBar(theme, context),
+            SizedBox(height: theme.spacingM),
+            _gradientAvatar(context),
+            AssetPortfolioHeading(title: selectedWallet.name, onTap: _onDropDownTapped, isCentered: true),
+            _textButtonRow(context, theme),
+            if (!selectedWallet.data.isBackedUp)
+              Padding(
+                padding: EdgeInsets.all(theme.spacingL),
+                child: CosmosWarningBox(
```
andrzejchm, line 54: you can now remove the `warning.png` icon, as its no longer used

starport_template/lib/stores/wallets_store.dart
```diff
@@ -97,7 +97,10 @@ class WalletsStore {
 
   final Observable<int?> _selectedWalletIndex = Observable(null);
 
-  set selectedWalletIndex(int? value) => Action(() => _selectedWalletIndex.value = value)();
+  set selectedWalletIndex(int? value) {
+    Action(() => _selectedWalletIndex.value = value)();
+    getBalances(selectedWallet.publicAddress);
```
andrzejchm, line 102: let's also check if the value has changed indeed and do the fetch only if it's different

starport_template/lib/widgets/transaction_history_list.dart
```diff
@@ -0,0 +1,55 @@
...
+  Widget build(BuildContext context) {
+    return CustomScrollView(
+      slivers: groupedHistory.keys.map(
+        (groupedMonthYear) {
+          final transactionsList = groupedHistory[groupedMonthYear] ?? [];
+          return SliverToBoxAdapter(
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+                TransactionHistoryGroupTitle(title: groupedMonthYear.split('-')[0]),
```
andrzejchm, line 36: like I've mentioned, we don't want to have any string splitting logic inside the view, so lets at least extract it to separate method, which name describes why this splitting is needed. Ideally you could create the method adjacent to the `formatDate` and give it descriptive name

andrzejchm (changes requested):
just 3 more comments

andrzejchm (approved):

PR #167, opened 2021-12-15
refactor: Add function for splitting grouped date key

wal33d006:


andrzejchm (commented):
not sure what `groupedKey` is and its a bit too specific to that single case. Let's call it `extractMonth` and later, we can make it more smart and accept different date formatting and extract from those aswell

andrzejchm (approved):

PR #178, opened 2021-12-17
refactor: Add app information extractor

wal33d006:


packages/cosmos_utils/lib/app_info_extractor.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:package_info/package_info.dart';
+
+Future<AppInfo> getAppInfo() async {
```
andrzejchm, line 3: awesome! well done on hiding the library behind the abstraction :)

andrzejchm (approved):

PR #181, opened 2022-01-03
refactor: Add app_info_provider.dart

wal33d006:
Created a provider which provides all the data from `AppInfo` separately. Now the `AppInfo` class is also completely abstracted behind this provider. Let me know if this is a good approach. Also, I made a property so that it doesn't fetch the information again if it's already filled up after the first fetch.


andrzejchm (approved):
awesome job!

PR #188, opened 2022-01-13
feat: add delete wallet to transaction_signing_gateway

AndreHaueisen:
Maybe I've missed it, but I could not find a way to delete the wallet from the device.
This PR is an attempt to solve that.


andrzejchm (approved):
awesome work! thanks for the contribution!

PR #190, opened 2022-01-18
refactor: Make title as nullable property and adjust height based on empty title

wal33d006:


packages/cosmos_ui_components/lib/components/cosmos_app_bar.dart
```diff
@@ -30,17 +30,18 @@ class CosmosAppBar extends StatelessWidget implements PreferredSizeWidget {
             ],
           ),
           const SizedBox(height: 20),
```
andrzejchm, line 32: you have to remove the SizedBox aswell, if the title is null, otherwise it will still overflow

andrzejchm (changes requested):

packages/cosmos_ui_components/lib/components/cosmos_app_bar.dart
wal33d006: I removed the sized box. Still it's giving an overflow for 11 pixels. I think it's due to `SafeArea`

packages/cosmos_ui_components/lib/components/cosmos_app_bar.dart
```diff
@@ -29,22 +31,24 @@ class CosmosAppBar extends StatelessWidget implements PreferredSizeWidget {
               if (actions != null) ...actions!,
             ],
           ),
-          const SizedBox(height: 20),
```
andrzejchm, line 32: you have now completely removed the SizedBox, which is needed if we have the title set. please add it back and just move it under the if statement:
```dart
if (title != null) 
  ...[
    SizedBox(...),
    Padding(child: Text(...),...),
  ],

andrzejchm (changes requested):

packages/cosmos_ui_components/lib/components/cosmos_app_bar.dart
wal33d006: Yup, missed that

andrzejchm (approved):

PR #204, opened 2022-01-31
fix: Fix cosmos_outline_button.dart

wal33d006:


packages/cosmos_ui_components/lib/components/cosmos_outline_button.dart
```diff
@@ -7,17 +7,26 @@ class CosmosOutlineButton extends StatelessWidget {
...
+    this.height = defaultHeight,
     this.suffixIcon,
   }) : super(key: key);
 
   final VoidCallback? onTap;
   final String text;
   final Widget? suffixIcon;
+  final double height;
+
+  static const defaultHeight = 50.0;
```
andrzejchm, line 19: instead of introducing exact copy of the same constant from `CosmosElevatedButton`, lets export this constants to a common place and use it in both places. this way whenever some1 changes the height, it will be changed for all buttons

andrzejchm (changes requested):

andrzejchm (approved):

PR #219, opened 2022-02-22
fix: (transaction) 🐛 fixed amount denom

### LINKED ISSUE #212: Change denom on the "Transferred!" page
**Desription**
Currently, the denom is hard-coded to `AKT`. Notice the same thing happens when the transaction is being transferred (the modal with a spinner).

**Expected behavior**
The denom should correspond to the one being sent.

Screenshot 2022-02-16 at 11 15 33

Zfinix:
Fixes #212

### DIFF
--- a/starport_template/lib/entities/amount.dart
+++ b/starport_template/lib/entities/amount.dart
@@ -4,6 +4,7 @@ class Amount {
   const Amount(this.value);
 
   Amount.fromString(String string) : value = Decimal.parse(string);
+  Amount.fromDouble(double val) : value = Decimal.parse('$val');
 
   Amount.fromInt(int int) : value = Decimal.fromInt(int);
 
@@ -19,6 +20,6 @@ extension StringAmount on String {
   Amount get amount => Amount.fromString(this);
 }
 
-extension IntAmount on int {
-  Amount get amount => Amount.fromInt(this);
+extension NumAmount on num {
+  Amount get amount => Amount.fromInt(toInt());
 }
--- a/starport_template/lib/entities/amount.dart
+++ b/starport_template/lib/entities/amount.dart
@@ -4,8 +4,6 @@ class Amount {
   const Amount(this.value);
 
   Amount.fromString(String string) : value = Decimal.parse(string);
-  Amount.fromDouble(double val) : value = Decimal.parse('$val');
-
   Amount.fromInt(int int) : value = Decimal.fromInt(int);
 
   final Decimal value;
@@ -20,6 +18,6 @@ extension StringAmount on String {
   Amount get amount => Amount.fromString(this);
 }
 
-extension NumAmount on num {
-  Amount get amount => Amount.fromInt(toInt());
+extension IntAmount on int {
+  Amount get amount => Amount.fromInt(this);
 }
--- a/starport_template/lib/pages/assets_transfer_sheet.dart
+++ b/starport_template/lib/pages/assets_transfer_sheet.dart
@@ -55,7 +55,7 @@ class _AssetsTransferSheetState extends State<AssetsTransferSheet> {
                   const ContentLoadingIndicator(),
                   const Spacer(),
                   Text(
-                    '${formatAmount(widget.recipientGetsAmountBalance.amount.value.toDouble(), symbol: '')} ${widget.recipientGetsAmountBalance.denom}'
+                    '${widget.recipientGetsAmountBalance.amount.value.toString()} ${widget.recipientGetsAmountBalance.denom}'
                         .toUpperCase(),
                     style: CosmosTextTheme.title0Medium,
                   ),
--- a/starport_template/lib/pages/assets_transfer_sheet.dart
+++ b/starport_template/lib/pages/assets_transfer_sheet.dart
@@ -1,17 +1,21 @@
 import 'package:cosmos_ui_components/cosmos_text_theme.dart';
 import 'package:cosmos_ui_components/cosmos_ui_components.dart';
+import 'package:cosmos_utils/amount_formatter.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_mobx/flutter_mobx.dart';
+import 'package:starport_template/entities/balance.dart';
 import 'package:starport_template/starport_app.dart';
 
 class AssetsTransferSheet extends StatefulWidget {
   const AssetsTransferSheet({
     required this.onTapDone,
+    required this.balance,
     Key? key,
   }) : super(key: key);
 
   final VoidCallback onTapDone;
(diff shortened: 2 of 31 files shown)


Zfinix:
> Fixes #212
Simulator Screen Shot - iPhone 13 Pro Max - 2022-02-25 at 16 22 18

starport_template/lib/pages/assets_transfer_sheet.dart
```diff
@@ -41,24 +47,43 @@ class _AssetsTransferSheetState extends State<AssetsTransferSheet> {
...
+                      height: 56,
+                      width: 56,
+                      child: Image.asset('assets/images/check_circle.png'),
+                    ),
                     SizedBox(height: theme.spacingXL),
                     Text('Transferred!', style: CosmosTextTheme.title2Bold),
                     const Spacer(),
-                    Text('1134.48 AKT', style: CosmosTextTheme.title0Medium),
+                    Text(
+                      '${widget.balance.amount.value.toDouble()} '
```
andrzejchm, line 82: same here

starport_template/lib/pages/assets_transfer_sheet.dart
```diff
@@ -41,24 +47,43 @@ class _AssetsTransferSheetState extends State<AssetsTransferSheet> {
...
+                  Text(
+                    'This may take up to 1 minute.',
+                    style: CosmosTextTheme.copy0Normal,
+                  ),
                   const Spacer(),
                   const ContentLoadingIndicator(),
                   const Spacer(),
-                  Text('1134.48 AKT', style: CosmosTextTheme.title0Medium),
+                  Text(
+                    '${formatAmount(widget.balance.amount.value.toDouble(), symbol: '')} ${widget.balance.denom}'
```
andrzejchm, line 58: we're loosing precision here, there is a reason why we went with `Decimal` objects instead of simply doubles. 

Here's short explanation: https://stackoverflow.com/a/3730040

you could try with:

```suggestion
                      '${widget.balance.amount.value.toString()}'
```
but if you need to control the precision, you can use `toStringAsPrecision()` then

cc: @wal33d006

starport_template/lib/pages/routing_page.dart
```diff
@@ -22,7 +22,6 @@ class _RoutingPageState extends State<RoutingPage> {
   void initState() {
     super.initState();
     initialize();
-    _loadWallets();
```
andrzejchm, line 25: are we sure we want to stop loading wallets on the app start?

starport_template/lib/pages/sign_transaction_page.dart
```diff
@@ -124,10 +125,14 @@ class _SignTransactionPageState extends State<SignTransactionPage> {
...
     showMaterialModalBottomSheet(
       context: context,
+      isDismissible: false, // Stop user from tapping outside sheet to dismiss
       backgroundColor: Colors.transparent,
       builder: (context) => SizedBox(
         height: MediaQuery.of(context).size.height / 2.24,
         child: AssetsTransferSheet(
+          balance: widget.balance.copyWith(
+            amount: Amount.fromDouble(recipientGetsAmount),
+          ),
```
andrzejchm, line 135: let's not manipulate data inside the widget's tree. You can create a getter for the `reicipientGetsAmount` balance:

```dart
Balance get recipientGetsAmountBalance => widget.balance.copyWith(
            amount: Amount.fromDouble(recipientGetsAmount),
          );
```

starport_template/lib/pages/assets_transfer_sheet.dart
```diff
@@ -1,25 +1,31 @@
...
 
 class AssetsTransferSheet extends StatefulWidget {
   const AssetsTransferSheet({
     required this.onTapDone,
+    required this.balance,
     Key? key,
   }) : super(key: key);
 
   final VoidCallback onTapDone;
+  final Balance balance;
```
andrzejchm, line 18: the name is a bit too ambiguous. how about:

```dart
final Balance recipientGetsAmountBalance
```

andrzejchm (changes requested):
solid work! I've left few changes proposals with explanation, if you have any questions, just drop me a message!

starport_template/lib/pages/sign_transaction_page.dart
wal33d006: Here we are using the `fromDouble` constructor for `Amount` which we shouldn't. The `recipientGetsAmount` is a decimal value coming from this line somewhere above this file
`double get recipientGetsAmount => widget.transaction.amount.value.toDouble() - widget.transaction.fee;`

Technically this manipulation shouldn't be inside the UI and secondly this `widget.transaction.fee` should be a type of `Amount` as well so that we can add/subtract within decimals. What do you think @andrzejchm ?

starport_template/lib/entities/amount.dart
```diff
@@ -19,6 +20,6 @@ extension StringAmount on String {
   Amount get amount => Amount.fromString(this);
 }
 
-extension IntAmount on int {
-  Amount get amount => Amount.fromInt(this);
+extension NumAmount on num {
```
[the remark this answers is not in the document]
Zfinix: You are right.

starport_template/lib/pages/sign_transaction_page.dart
andrzejchm: good catch @wal33d006 ! definetely we should stick with `Amount/Decimal` and not convert back and forth from/to doubles

starport_template/lib/pages/routing_page.dart
Zfinix: Yes i discovered this s the cause of multiple account balances being loaded we already call `_loadWallets()` in `initialize()`

starport_template/lib/pages/assets_transfer_sheet.dart
```diff
@@ -41,24 +46,43 @@ class _AssetsTransferSheetState extends State<AssetsTransferSheet> {
...
+                      height: 56,
+                      width: 56,
+                      child: Image.asset('assets/images/check_circle.png'),
+                    ),
                     SizedBox(height: theme.spacingXL),
                     Text('Transferred!', style: CosmosTextTheme.title2Bold),
                     const Spacer(),
-                    Text('1134.48 AKT', style: CosmosTextTheme.title0Medium),
+                    Text(
+                      '${widget.recipientGetsAmountBalance.amount.value.toDouble()} '
```
wal33d006, line 81: `formatAmount(widget.recipientGetsAmountBalance.amount.value.toDouble())`

wal33d006:
Looks amazing! Let Andrzej have a look and we are good to go!

andrzejchm (approved):
awesome work!

PR #226, opened 2022-03-03
refactor: Rename wallet to account

wal33d006:
Big PR, nothing fancy just renamed everything that was possible.


Zfinix (approved):
LGTM 🙌🏾

andrzejchm (commented):
good job! Since this is a public package used by other developers, we need to provide a migration guide for them, let's add a changelog file with small section of migration guide for versions older than 2.0.3 stating all the classes that have changed their names

wal33d006:
Updated `CHANGELOG.md` file.
Followed [this](https://github.com/olivierlacan/keep-a-changelog/blob/main/CHANGELOG.md) convention (open to suggestions if that's not a very cool way).

Added previous change logs by simply picking up points from releases, most of which are not related to this package. Let me know if we need to remove them and add something else entirely.

@andrzejchm

Issue #203, opened 2022-01-31
[Feature] return full transaction


andrzejchm:
by default, `broadcastTransaction` method returns `TransactionHash` if the transaction is successfully broadcasted to the chain, but it would be also beneficial to get the full transaction object that consists of the full logs and any potential errors that would explain why transaction has failed.

PR #231, opened 2022-03-09
refactor: return TransactionResponse instead of TransactionHash after successful broadcast

Zfinix:
Closes https://github.com/tendermint/flutter/issues/203

### DIFF
--- a/packages/transaction_signing_gateway/lib/model/transaction_log.dart
+++ b/packages/transaction_signing_gateway/lib/model/transaction_log.dart
@@ -4,7 +4,7 @@ class TransactionLog extends Equatable {
   const TransactionLog({
     required this.msgIndex,
     required this.log,
-    this.events,
+    this.events = const [],
   });
 
   final int msgIndex;
--- a/packages/transaction_signing_gateway/lib/model/transaction_log.dart
+++ b/packages/transaction_signing_gateway/lib/model/transaction_log.dart
@@ -12,7 +12,11 @@ class TransactionLog extends Equatable {
   final List<Object?>? events;
 
   @override
-  List<Object> get props => [msgIndex, log ?? '', events ?? []];
+  List<Object?> get props => [
+    msgIndex, 
+    log, 
+    events,
+];
 
   TransactionLog copyWith({
     int? msgIndex,
--- a/packages/transaction_signing_gateway/lib/model/transaction_log.dart
+++ b/packages/transaction_signing_gateway/lib/model/transaction_log.dart
@@ -13,10 +13,10 @@ class TransactionLog extends Equatable {
 
   @override
   List<Object?> get props => [
-    msgIndex, 
-    log, 
-    events,
-];
+        msgIndex,
+        log,
+        events,
+      ];
 
   TransactionLog copyWith({
     int? msgIndex,
--- a/packages/transaction_signing_gateway/lib/model/transaction_log.dart
+++ b/packages/transaction_signing_gateway/lib/model/transaction_log.dart
@@ -1,0 +1,28 @@
+import 'package:equatable/equatable.dart';
+
+class TransactionLog extends Equatable {
+  const TransactionLog({
+    required this.msgIndex,
+    required this.log,
+    this.events,
+  });
+
+  final int msgIndex;
+  final String? log;
+  final List<Object?>? events;
+
+  @override
+  List<Object> get props => [msgIndex, log ?? '', events ?? []];
+
+  TransactionLog copyWith({
+    int? msgIndex,
+    String? log,
+    List<Object?>? events,
+  }) {
+    return TransactionLog(
+      msgIndex: msgIndex ?? this.msgIndex,
+      log: log ?? this.log,
+      events: events ?? this.events,
+    );
+  }
+}
--- a/packages/transaction_signing_gateway/lib/model/transaction_log.dart
+++ b/packages/transaction_signing_gateway/lib/model/transaction_log.dart
@@ -8,8 +8,8 @@ class TransactionLog extends Equatable {
   });
 
   final int msgIndex;
-  final String? log;
-  final List<Object?>? events;
+  final String log;
+  final List<Object?> events;
 
   @override
   List<Object?> get props => [
--- a/packages/transaction_signing_gateway/lib/model/transaction_response.dart
+++ b/packages/transaction_signing_gateway/lib/model/transaction_response.dart
@@ -1,0 +1,128 @@
+import 'package:alan/alan.dart';
+import 'package:decimal/decimal.dart';
+import 'package:equatable/equatable.dart';
+
+class TransactionResponse extends Equatable {
+  const TransactionResponse({
+    required this.hash,
+    required this.data,
+    required this.logs,
+    required this.rawLog,
+    required this.gasUsed,
+    required this.gasWanted,
+    required this.timestamp,
+  });
+
+  factory TransactionResponse.fromTxResponse(TxResponse tx) {
+    return TransactionResponse(
+      hash: TransactionHash(hash: tx.txhash),
+      data: tx.data,
+      rawLog: tx.rawLog,
+      logs: List<TransactionLog>.from(
+        tx.logs.map(
+          (log) => TransactionLog(
+            log: log.log,
+            msgIndex: log.msgIndex,
+            events: List.from(
+              log.events.map(
+                (e) => e.toProto3Json(),
+              ),
+            ),
+          ),
+        ),
+      ),
+      gasUsed: Decimal.parse('${tx.gasUsed}'),
+      gasWanted: Decimal.parse('${tx.gasWanted}'),
+      timestamp: DateTime.fromMillisecondsSinceEpoch(int.parse(tx.timestamp)),
+    );
+  }
+
+  final TransactionHash hash;
+  final String data;
+  final String rawLog;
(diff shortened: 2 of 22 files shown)


packages/transaction_signing_gateway/lib/transaction_broadcaster.dart
```diff
@@ -1,11 +1,11 @@
+import 'package:alan/alan.dart';
 import 'package:dartz/dartz.dart';
 import 'package:transaction_signing_gateway/model/private_account_credentials.dart';
 import 'package:transaction_signing_gateway/model/signed_transaction.dart';
 import 'package:transaction_signing_gateway/model/transaction_broadcasting_failure.dart';
-import 'package:transaction_signing_gateway/model/transaction_hash.dart';
 
 abstract class TransactionBroadcaster {
-  Future<Either<TransactionBroadcastingFailure, TransactionHash>> broadcast({
+  Future<Either<TransactionBroadcastingFailure, TxResponse>> broadcast({
```
wal33d006, line 8: Why are we needing to return a class of the type from inside the generated files? Is there a problem in returning a type of our own class?
Also, this is affecting the abstraction as well. Technically, the abstraction shouldn't return a very specific type (and that too from the generated files).
I don't know the context of the bug, but do we have a strong reason of doing this?
cc: @andrzejchm

wal33d006: I just read the bug description, we should allow the previous class to have all the properties of the full transaction object, rather than returning the object of a generated file. Technically refactoring `TransactionHash` to `TransactionResponse` of `Transaction` simply and keeping all the parameters required there.

Zfinix: sure we can do that

andrzejchm: I agree with @wal33d006 we should wrap the Transaction into our own class with all the relevant fields copied over. Like:

```dart
class TransactionResponse extends Equatable {
  final TransactionHash hash;
  final String data;
  final List<TransactionLog> logs;
  final Decimal gasUsed;
  final Decimal gasWanted;
  final DateTime timestamp;
}
```

andrzejchm (changes requested):

packages/transaction_signing_gateway/lib/transaction_broadcaster.dart
Zfinix: done this and will soon push it

Zfinix: I created a factor constructor that takes in `TxResponse` and spits out  `TransactionResponse` 

```dart
 factory TransactionResponse.fromTxResponse(TxResponse tx) {
    return TransactionResponse(
      hash: TransactionHash(hash: tx.txhash),
      data: tx.data,
      rawLog: tx.rawLog,
      logs: List<TransactionLog>.from(
        tx.logs.map(
          (log) => TransactionLog(
            log: log.log,
            msgIndex: log.msgIndex,
            events: List.from(
              log.events.map(
                (e) => e.toProto3Json(),
              ),
            ),
          ),
        ),
      ),
      gasUsed: Decimal.parse('${tx.gasUsed}'),
      gasWanted: Decimal.parse('${tx.gasWanted}'),
      timestamp: DateTime.fromMillisecondsSinceEpoch(int.parse(tx.timestamp)),
    );
  }


```

Zfinix:
I created a factor constructor that takes in TxResponse and spits out TransactionResponse

```dart
 factory TransactionResponse.fromTxResponse(TxResponse tx) {
    return TransactionResponse(
      hash: TransactionHash(hash: tx.txhash),
      data: tx.data,
      rawLog: tx.rawLog,
      logs: List<TransactionLog>.from(
        tx.logs.map(
          (log) => TransactionLog(
            log: log.log,
            msgIndex: log.msgIndex,
            events: List.from(
              log.events.map(
                (e) => e.toProto3Json(),
              ),
            ),
          ),
        ),
      ),
      gasUsed: Decimal.parse('${tx.gasUsed}'),
      gasWanted: Decimal.parse('${tx.gasWanted}'),
      timestamp: DateTime.fromMillisecondsSinceEpoch(int.parse(tx.timestamp)),
    );
  }
```

packages/transaction_signing_gateway/lib/model/transaction_response.dart
```diff
@@ -0,0 +1,140 @@
...
+      info: info ?? this.info,
+      logs: logs ?? this.logs,
+      gasUsed: gasUsed ?? this.gasUsed,
+      gasWanted: gasWanted ?? this.gasWanted,
+      timestamp: timestamp ?? this.timestamp,
+    );
+  }
+}
+
+class TransactionHash extends Equatable {
```
andrzejchm, line 94: lets not mix two classes into single file. All entities should have their own file

packages/transaction_signing_gateway/pubspec.yaml
```diff
@@ -17,6 +17,7 @@ dependencies:
   encrypt: ^5.0.1
   alan: ^0.40.2+1
   uuid: ^3.0.5
+  decimal: 1.5.0
```
andrzejchm, line 20: since transaction_signing_gateway is a public package, let's not be too strict about versioning of the libraries it uses, otherwise its going to cause a lot of headache for dependency resolving in the client's code.
```suggestion
  decimal: ^1.5.0
```

starport_template/lib/utils/token_sender.dart
```diff
@@ -49,7 +49,7 @@ class TokenSender {
         );
     result.fold(
       (fail) => throw fail as Object,
-      (hash) => debugPrint('new TX hash: $hash'),
+      (tx) => debugPrint('new TX hash: ${tx.toString()}'),
```
andrzejchm, line 52:
```suggestion
      (tx) => debugPrint('new TX hash: $tx'),
```

packages/transaction_signing_gateway/lib/model/transaction_response.dart
```diff
@@ -0,0 +1,140 @@
...
+    required this.logs,
+    required this.info,
+    required this.codespace,
+    required this.rawLog,
+    required this.gasUsed,
+    required this.gasWanted,
+    required this.timestamp,
+  });
+
+  factory TransactionResponse.fromTxResponse(TxResponse tx) {
```
andrzejchm, line 18: let's try to always keep the models agnostic of the underlying libraries, thus the factory that we have here we could add a separate file `model_translations` and add this as an extension:

```dart
extension TxResponseTranslator on TxResponse {
TransactionResponse toTransactionResponse() => TransactionResponse( ... );
```
this is because we might want to support other chains later in the future (like Ethereum) and we want to adhere to the [Open-Closed Principle](https://deviq.com/principles/open-closed-principle)

andrzejchm (changes requested):
awesome work! Just few minor comments left

packages/transaction_signing_gateway/pubspec.yaml
Zfinix: Ahh noted

packages/transaction_signing_gateway/lib/model/transaction_response.dart
Zfinix: Added fixes for this

packages/transaction_signing_gateway/pubspec.yaml
Zfinix: Added fix

packages/transaction_signing_gateway/lib/model/transaction_response.dart
Zfinix: Added fix

packages/transaction_signing_gateway/lib/model/transaction_log.dart
```diff
@@ -0,0 +1,28 @@
...
+class TransactionLog extends Equatable {
+  const TransactionLog({
+    required this.msgIndex,
+    required this.log,
+    this.events,
+  });
+
+  final int msgIndex;
+  final String? log;
+  final List<Object?>? events;
```
andrzejchm, line 12: lets prefer non-nullable types in our own models, so:
```suggestion
  final String log;
  final List<Object?> events;
```

packages/transaction_signing_gateway/lib/model/transaction_log.dart
```diff
@@ -0,0 +1,28 @@
...
+
+  @override
+  List<Object> get props => [msgIndex, log ?? '', events ?? []];
```
andrzejchm, line 15:
```suggestion
  List<Object?> get props => [
    msgIndex, 
    log, 
    events,
];
```

andrzejchm (approved):
2 more ultra minors and we are good to go

andrzejchm (approved):
👏🏻

packages/transaction_signing_gateway/lib/transaction_broadcaster.dart
Zfinix: Hi @wal33d006 waiting for your review

Zfinix:
> 👏🏻

Can I merge this in before creating a new PR

andrzejchm:
Ive merged it, @wal33d006 should you have any comments on the pr, feel free to drop them here and we can work on potential improvements in a separate pr

PR #244, opened 2022-03-29
feat: created config base env file

Zfinix:


fadeev:
This looks good for a small config, but would it scale? For example, one of the things we'll add is fee configuration (see Keplr for example). It might look something like this:

```yml
fees:
  - denom: uosmo
    amount:
      low: 1000
      normal: 3000
      fast: 4000
  - denom: uion
    amount:
      low: 10
      normal: 20
      fast: 40
```

Zfinix:
> This looks good for a small config, but would it scale? For example, one of the things we'll add is fee configuration (see Keplr for example). It might look something like this:
> 
> ```yaml
> fees:
>   - denom: uosmo
>     amount:
>       low: 1000
>       normal: 3000
>       fast: 4000
>   - denom: uion
>     amount:
>       low: 10
>       normal: 20
>       fast: 40
> ```
> 
> Here's the amount of configuration Keplr requires:
> 
> https://github.com/chainapsis/keplr-example/blob/c3d6b483d74aa1521f13d40b291fd62361b2c518/src/main.js#L28-L113

I believe it would scale, the problem here is that we don't want to build a parser for that structure and we want to be able to still use `--dart-define` and also change the env variables using hotreload and hot restart, this is only possible if we  write it in `dart`

starport_template/lib/main.dart
```diff
@@ -16,7 +17,8 @@ void main() {
 }
 
 void _buildDependencies() {
-  StarportApp.baseEnv = BaseEnv();
+  final env = BaseEnv();
+  StarportApp.baseEnv = BaseEnvUtil(env);
```
andrzejchm, line 21: let's just get rid of the `String.fromEnvironment` reading as it causes too many problems.
```suggestion
  StarportApp.appConfig = AppConfig();
```

andrzejchm (changes requested):

starport_template/lib/base_env.dart
```diff
@@ -0,0 +1,43 @@
+class BaseEnv {
```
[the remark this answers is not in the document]
Zfinix: yes i agree

starport_template/lib/utils/env_util.dart
```diff
@@ -0,0 +1,82 @@
+// ignore_for_file: do_not_use_environment
+
+import 'dart:io';
+
+import 'package:alan/alan.dart';
+import 'package:grpc/grpc.dart';
+import 'package:starport_template/base_env.dart';
+
+class BaseEnvUtil {
```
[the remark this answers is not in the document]
Zfinix: I have removed the BaseEnvUtil class enirely

starport_template/lib/utils/env_util.dart
```diff
@@ -0,0 +1,82 @@
...
+            host: grpcUrl ?? env.envGrpcUrl,
+            port: int.parse(grpcPort ?? env.envGrpcPort),
+            credentials: int.parse(grpcPort ?? env.envGrpcPort) == 443
+                ? const ChannelCredentials.secure()
+                : const ChannelCredentials.insecure(),
+          ),
+        ),
+        baseApiUrl = lcdUrl != null && lcdPort != null ? '$lcdUrl:$lcdPort' : env.envBaseApiUrl;
+
+  final NetworkInfo networkInfo;
```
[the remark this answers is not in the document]
Zfinix: I have added it to AppConfig

starport_template/lib/main.dart
Zfinix: I have removed the  `String.fromEnvironment`

Zfinix (commented):
@andrzejchm  i also added updates to the docs

starport_template/lib/app_config.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:alan/alan.dart';
+import 'package:grpc/grpc.dart';
+
+class AppConfig {
+  AppConfig({
+    this.lcdUrl = 'http://10.0.2.2',
```
andrzejchm, line 6: - `10.0.2.2` - localhost for android emulators
- `lhttp://localhost` - ios simulator

you'll have to stick with only 1 version, I suggest going with `http://localhost` for both LCD and grpc

andrzejchm (approved):
awesome job @Zfinix ! Its much cleaner now. one comment and we're good to merge

wal33d006 (approved):

starport_template/lib/app_config.dart
Zfinix: Okay

Zfinix:
This is ready to be merged guys @wal33d006 @andrzejchm

andrzejchm:
please merge at your convenience as soon as the CI checks pass :)

Zfinix:
Merging this now

PR #246, opened 2022-03-31
fix: added formatting to sign_transaction_tab

### LINKED ISSUE #225: Fix Overflow and Amount Formatting on Sign Transaction Page
**Description**
There is an overflow of text when a very large number is inputted into the send transaction amount field

**To Reproduce**
1.  Open the app
2. Click the "Send" button
3. Select Asset
4. Enter Receiver's address and 10,000,000 as the amount

**Expected behavior**
1.
image

Zfinix:
This closes  #225 and https://github.com/tendermint/flutter/pull/246
image


starport_template/pubspec.yaml
```diff
@@ -24,6 +24,7 @@ dependencies:
   intl: 0.17.0
   alan: 0.40.2+1
   shared_preferences: 2.0.12
+  flutter_masked_text2: ^0.9.1
```
andrzejchm, line 27: CleanShot 2022-03-31 at 13 41 08@2x
 this is exactly the reason why we refrain from putting too many libraries into the template, let's not impose those dependencies on end users and keep the template as simple as possible.
 
I believe masked text is not a core feature of the template and can be  easly omitted

andrzejchm: also, this looks like its not used anywhere

starport_template/lib/widgets/sign_transaction_tab_view_item.dart
```diff
@@ -21,14 +22,23 @@ class SignTransactionTabViewItem extends StatelessWidget {
...
+          Row(
+            children: [
+              Text(text, style: CosmosTextTheme.titleS),
+              const Spacer(),
+              CosmosTokenAvatar(text: balance.denom.text),
+            ],
+          ),
+          SizedBox(width: theme.spacingL),
+          Text(
+            '${NumberFormat.decimalPercentPattern(decimalDigits: 5).format(amount)} ${balance.denom.text.toUpperCase()}',
```
andrzejchm, line 38: we're introducing yet another formatting of the amount. let's extract formatting of the amount to the`Balance` class

andrzejchm (changes requested):
this does not look good for larger amounts
CleanShot 2022-03-31 at 13 54 45@2x

starport_template/pubspec.yaml
Zfinix: Yeah i was going to remove it

starport_template/lib/widgets/sign_transaction_tab_view_item.dart
andrzejchm: formatting amount is rather easy task and can differ between emeris and template and other apps, so no need to put it in cosmos_utils. its enough to have a helper method inside either the `Balance` class or `Denom` class to format it, currently, in Emeris we use the following:

```dart
String amountWithDenomText(Amount amount) => '${formatAmount(amount.value.toDouble(), symbol: '')} $displayName';
```
inside the `Denom` class, and the `Balance class has the same method that proxies to `Denom`:

```dart
String get amountWithDenomText => denom.amountWithDenomText(amount);
```

Zfinix: Great

wal33d006 (approved):
Good to go after the change requests are resolved

starport_template/lib/widgets/sign_transaction_tab_view_item.dart
Zfinix: image
@andrzejchm this gives you this

Zfinix:
This is ready to be merged @andrzejchm

andrzejchm:
@Zfinix best way to request additional review is to to it in the `reviewers` section, this way I'll have it on my list of prs to review and its much bigger chance it won't get lost in the plethora of github emails :)
CleanShot 2022-04-13 at 20 49 19@2x

Zfinix:
Ahh i see

packages/cosmos_ui_components/pubspec.lock
```diff
@@ -69,7 +69,7 @@ packages:
     description:
       path: "packages/cosmos_utils"
```
andrzejchm, line 70: we should not remove all the `.fvm/fvm_config.json` files. those are needed to make sure we're all using the same version of flutter. if its outdated and differs from the one in `pull_request.yaml` then let's update those

andrzejchm (changes requested):
one last thing

packages/cosmos_ui_components/pubspec.lock
Zfinix: Sure

Zfinix:
Moved to https://github.com/tendermint/flutter/pull/259

PR #260, opened 2022-04-18
refactor: Add config for generating assets using `build_runner` and `flutter_gen`

wal33d006:


Zfinix:
Is there any reason we are adding build_runner and flutter_gen at the same time, I actually feel like we shouldn't do this as it forces flutter to add all the assets at runtime, while including the complexity of running build_runner each time in the background or any time we add a new asset, Sadly i fail to see the advantages

wal33d006:
According to the documentation of flutter_gen, we need to do that, and since we are shifting to generated assets instead if quoted strings, this would be required. This was actually something that was intended for the product, I thought I should add it in the shared code as well for consistency. We can leave it like this for shared code if we want. 
cc: @andrzejchm 

> Is there any reason we are adding build_runner and flutter_gen at the same time, I actually feel like we shouldn't do this as it forces flutter to add all the assets at runtime, while including the complexity of running build_runner each time in the background or any time we add a new asset, Sadly i fail to see the advantages

andrzejchm (changes requested):
we should not add flutter_gen to the public repo. the issue was meant only for the private repo

Zfinix (approved):
Yeah i believe we should leave out build runner for now

wal33d006:
I think we should abandon this PR

PR #262, opened 2022-04-21
refactor: Add casta font files and text style

wal33d006:


packages/cosmos_ui_components/lib/cosmos_text_theme.dart
```diff
@@ -4,6 +4,7 @@ abstract class CosmosTextTheme {
   factory CosmosTextTheme._() => throw UnsupportedError('Not meant to be instantiated');
 
   static const String fontInter = 'Inter';
+  static const String fontCasta = 'Casta';
```
wal33d006, line 7: @andrzejchm 
Can't figure out how two separate font families should be handled, for now just added a new style with this new font family as a parameter to the property.

Zfinix (approved):

packages/cosmos_ui_components/lib/cosmos_text_theme.dart
Zfinix: Oh this is easy you just define a new family in the assets section of pubspec

wal33d006: Yes I already did that, I am asking about injecting the family for different fonts dynamically

Zfinix: 🤔 I don't understand

andrzejchm: @wal33d006 this is a copyrighted font that should not be included in the public repo, its specific to the emeris app and should be placed there

Zfinix: Oh yeah thats a draw back

PR #265, opened 2022-04-29
feat: Migrating linting to cosmos_lint package

Zfinix:
This ensures that all packages are dependent on `cosmos_lint` and removes all instances of `lint`


andrzejchm (approved):
awesome, I love it!

andrzejchm (approved):

PR #276, opened 2022-05-12
refactor: Update custom switch to have Cosmos specific gradient

wal33d006:
I have added this custom switch (created it from scratch, as per the design, with animations, gradient, and the custom toggle pin). Do let me know if this is a good idea. It apparently works well and has no issues with both theme modes.

Light:
Screenshot 2022-05-12 at 6 49 35 PM
Screenshot 2022-05-12 at 6 49 39 PM

Dark:
Screenshot 2022-05-12 at 6 49 20 PM
Screenshot 2022-05-12 at 6 49 15 PM


packages/cosmos_ui_components/lib/components/cosmos_switch.dart
```diff
@@ -13,26 +14,89 @@ class CosmosSwitch extends StatelessWidget {
...
-            activeColor: theme.colors.link,
-            thumbColor: checked ? theme.colors.background : theme.colors.text,
-            trackColor: theme.colors.background,
+          scale: 1.1,
+          child: Align(
+            alignment: animation.value,
+            child: SizedBox(
+              width: 30,
+              height: 30,
+              child: Image.asset('assets/images/toggle.png'),
```
andrzejchm, line 98: let's not hardcode the asset here, since this component is reused in both starport template and emeris (and potentially other apps) we might want to customize the toggle
```suggestion
              child: Image.asset(toggleAsset),
```

packages/cosmos_ui_components/lib/components/cosmos_switch.dart
```diff
@@ -13,26 +14,89 @@ class CosmosSwitch extends StatelessWidget {
...
-          borderRadius: BorderRadius.circular(1000),
-          border: Border.all(
-            color: checked ? Colors.transparent : theme.colors.text.withOpacity(0.1),
-          ),
+          borderRadius: BorderRadius.circular(30),
+          gradient: checked
+              ? const LinearGradient(
+                  stops: [0, 0.5866, 1],
+                  colors: GlobalConstants.cosmosGradientColors,
+                )
```
andrzejchm, line 87: let's parametrize those values:
```suggestion
              ? const LinearGradient(
                  stops: checkedGradientStops,
                  colors: gradientColors,
                )
```

packages/cosmos_ui_components/lib/components/cosmos_switch.dart
```diff
@@ -13,26 +14,89 @@ class CosmosSwitch extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return GestureDetector(
+      onTap: () {
+        if (_animationController.isCompleted) {
+          _animationController.reverse();
+        } else {
+          _animationController.forward();
+        }
```
andrzejchm, line 56: this is logic inside build method. also, specifying ANimationController is too low level and requires a lot of boilerplate code and maintenance. you have widgets for such simple animations, for example `TweenAnimationBuilder`. Have a look at `expandable_container.dart` for an example how to use it

andrzejchm (changes requested):

packages/cosmos_ui_components/lib/components/cosmos_switch.dart
```diff
@@ -1,101 +1,59 @@
...
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
-class CosmosSwitch extends StatefulWidget {
+class CosmosSwitch extends StatelessWidget {
   const CosmosSwitch({
     required this.checked,
+    required this.toggleAsset,
+    required this.checkedGradientStops,
+    required this.gradientColors,
```
andrzejchm, line 11: by "parametrizing" i meant to make the values optional. right now its cumbersome to use `CosmosSwitch` as it requires a lot of parameters and it does not work "out of the box". 
```suggestion
   this.checkedGradientStops = [ ... ], 
   this.gradientColors = GlobalConstants.cosmosGradientColors,
```

So a rule of a thumb:
Always provide defaults for the non-essential params for the widgets so that the widget's consumer can easily create widgets with minimal setup, but allow them to alter the values however they want, if needed

packages/cosmos_ui_components/lib/components/cosmos_switch.dart
```diff
@@ -1,101 +1,59 @@
...
-      begin: widget.checked ? Alignment.centerRight : Alignment.centerLeft,
-      end: widget.checked ? Alignment.centerLeft : Alignment.centerRight,
-    ).animate(
-      CurvedAnimation(
-        parent: _animationController,
-        curve: Curves.linear,
-      ),
-    );
-  }
+  final String toggleAsset;
```
andrzejchm, line 18: you've now removed the `toggle.png` which makes the asset `required` and will make the starport template and emeris app not compiling. the param should be optional and have default value of `toggle.png` that should be embedded into the library

andrzejchm (changes requested):

packages/cosmos_ui_components/lib/components/cosmos_switch.dart
```diff
@@ -1,38 +1,63 @@
...
     this.onChanged,
     Key? key,
   }) : super(key: key);
 
   final bool checked;
   final ValueChanged<bool>? onChanged;
+  final String toggleAsset;
+  final List<double> checkedGradientStops;
+  final List<Color> gradientColors;
+  final String? package;
```
andrzejchm, line 23: can you confirm this is working in starport template or emeris without setting up `package`? otherwise we need to update it to:
```suggestion
  final String? package = packageName;
```

wal33d006: There is a little bit of an issue here.

If we assign `packageName` as a default value, then whatever the consumer wants to add an asset of their own, they will have to pass null inside `packageName` parameter specifically, otherwise `CosmosSwitch` will start finding the consumer side asset inside its own assets due to defaulted `packageName`.

But I think in our case this is the only thing we can do.

andrzejchm: so maybe we could default toggleAsset to null, and then inside the widget do the following:

```dart
Image.asset(
   imageAsset ?? 'assets/images/toggle.png',
   package: package ?? (imageAsset == null ? packageName : null),
 )
 ```

wal33d006: This would work I guess. Thanks

andrzejchm (approved):

PR #278, opened 2022-05-13
fix: [TxResponseTranslator extension] Fix parse timestamp property

mduccc:
Use `int.tryParse(timestamp)` instead `int.parse(timestamp)`.

`int.parse(timestamp)` often throw an exception.


andrzejchm (approved):
awesome! totally makes sense

andrzejchm:
@mduccc could you fix code formatting in your pr using `flutter format -l 120 ` command?

mduccc:
> awesome! totally makes sense

Yeah. I complete fixing code formatting with that command

PR #279, opened 2022-05-13
fix: Fix disabled style for `CosmosElevatedButton`

wal33d006:
`CosmosElevatedButton` was picking up disabled style from Flutter's theme because it uses onTap nullability to disable the button, which was causing the button to disappear in case of dark theme. Just fixed that.

### DIFF
--- a/packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
+++ b/packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
@@ -34,7 +34,7 @@ class CosmosElevatedButton extends StatelessWidget {
   Widget build(BuildContext context) {
     final theme = CosmosTheme.of(context);
     return ElevatedButton(
-      onPressed: onTap ?? () {},
+      onPressed: onTap,
       style: onTap == null ? _disabledStyle(context) : _enabledStyle(context),
       child: Row(
         mainAxisSize: MainAxisSize.min,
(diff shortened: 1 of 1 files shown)


packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
```diff
@@ -34,14 +34,15 @@ class CosmosElevatedButton extends StatelessWidget {
   Widget build(BuildContext context) {
     final theme = CosmosTheme.of(context);
     return ElevatedButton(
-      onPressed: onTap,
+      onPressed: onTap ?? () {},
```
andrzejchm, line 37: can't we simply update the "disabled" colors for the button? instead of all those  if/elses ?
i.e:
```dart
ElevatedButton(
    style: onTap == null ? _disabledStyle() : _enabledStyle(),
    ...
)
```

andrzejchm (changes requested):

packages/cosmos_ui_components/lib/components/cosmos_elevated_button.dart
wal33d006: Yes we can do that, still we would have to provide an empty function to `onTap`, because if `onTap` is `null`, Flutter's internal code automatically applies disabled style and the button disappears on dark mode. Therefore, in case of a `null` `onTap` I had to provide it an empty function and manually remove the splash effect and shadow in order to make it feel like it's not tappable.

andrzejchm: you still should be able to alternate the way flutter displays disabled button: https://www.flutterbeads.com/elevated-button-color-in-flutter/#Customize-the-Elevated-Button-Color-for-Disable-State

andrzejchm (approved):
super

PR #281, opened 2022-05-16
feat: Dark theme

wal33d006:
The Ignite template now supports dark theme.
There is a toggle inside settings which allows you to change theme. Right now the updated theme is not saved in preferences which is why if you selected a dark theme you won't get it back after you reopen the app.
This will be implemented in a separate PR after this one is merged.
Simulator Screen Shot - iPhone 12 Pro - 2022-05-16 at 16 25 36
Simulator Screen Shot - iPhone 12 Pro - 2022-05-16 at 16 26 22


Zfinix:
Awesomeeeee work👍👍👍

Zfinix (approved):
Great work @wal33d006

starport_template/lib/starport_app.dart
```diff
@@ -22,16 +24,18 @@ class StarportApp extends StatelessWidget {
...
   Widget build(BuildContext context) {
-    return CosmosTheme(
-      child: MaterialApp(
-        title: 'Starport template',
-        //darkTheme: CosmosTheme.buildDarkAppTheme(), // enable for dark mode
-        theme: CosmosTheme.buildAppTheme(),
-        home: const RoutingPage(),
+    return Observer(
+      builder: (context) => CosmosTheme(
+        themeData: themeStore.isDarkTheme ? CosmosTheme.darkThemeData : const CosmosThemeData(),
```
andrzejchm, line 34: could we implement the suggestion from this comment first:
https://github.com/tendermint/flutter/pull/266#issuecomment-1118337803

Also, we could get rid of this discrepancy of having light theme in default constructor and dark theme in static field. instead, lets have separate constructors for those:
```dart
CosmosThemeData.light()
CosmosThemeData.dark()
```

then, we could specify this piece as:

```dart
Observer(
    builder: (context) => CosmosTheme(
        brightness: () => themeStore.isDarkTheme ? Brightness.dark : Brightness.light,
        // by default, CosmosTheme would have default params set up for light and dark theme, so we don't need to provide anything here
    )
)
```

and the `CosmosTheme` constructor:
```dart
const CosmosTheme({
    this.themeData = CosmosThemeData.light(),
    this.darkThemeData = CosmosThemeData.dark(),
    ...
})

andrzejchm (changes requested):

andrzejchm (approved):

PR #291, opened 2022-05-18
fix: Remove unused function. Fix ternary condition in `CosmosTheme`

wal33d006:


andrzejchm (approved):
love it! thanks for your patience 🙏🏻

PR #293, opened 2022-05-18
fix: Remove input decoration theme from theme_converter.dart

wal33d006:
The `InputDecorationTheme` needed to be removed, as we only need some of the Flutter's theme properties and not all of them. In this case, `InputDecorationTheme` was implicitly being used by `PinPut` which showed our passcode screen like the image below since it used `TextField`s which use `InputDecorationTheme`. We will continue to see what properties to remove if they cause issues in the future as well.
simulator_screenshot_6EC299AE-CD34-4EC7-89E2-B3BE61387FB5


andrzejchm (approved):
Will it work for dark theme/light theme now?

wal33d006:
> Will it work for dark theme/light theme now?

Yes it would. We tested it before. When giving inout decoration from Flutter theme, we had to wrap `PinPut` in a `Theme` widget in order to override. Since we are now using custom input decoration in our `CosmosTextField` directly, this can be removed and if any consume app wants to override Flutter theme's input decoration, they easily can because it's already default.

PR #297, opened 2022-05-19
Feat: create Error screen

Zfinix:
I have created a default error screen using flutter's inbuilt error widget and i have also passed the error from account creation to it

### DIFF
--- a/starport_template/lib/starport_app.dart
+++ b/starport_template/lib/starport_app.dart
@@ -1,6 +1,6 @@
 import 'package:alan/alan.dart';
 import 'package:cosmos_auth/auth/cosmos_auth.dart';
-import 'package:cosmos_ui_components/cosmos_theme.dart';
+import 'package:cosmos_ui_components/cosmos_ui_components.dart';
 import 'package:cosmos_utils/cosmos_utils.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_mobx/flutter_mobx.dart';
@@ -34,6 +34,7 @@ class StarportApp extends StatelessWidget {
         brightness: themeStore.isDarkTheme ? Brightness.dark : Brightness.light,
         child: Builder(
           builder: (context) {
+            setErrorBuilder(context);
             return MaterialApp(
               title: 'Starport template',
               theme: CosmosTheme.of(context).buildFlutterTheme(),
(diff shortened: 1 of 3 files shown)


wal33d006:
Looks good to me, can we make it support `CosmosTheme`? Currently the background color and text colors are all fixed.

wal33d006 (changes requested):
Looks good to me, can we make it support `CosmosTheme`? Currently the background color and text colors are all fixed.

starport_template/lib/starport_app.dart
```diff
@@ -34,6 +34,7 @@ class StarportApp extends StatelessWidget {
         brightness: themeStore.isDarkTheme ? Brightness.dark : Brightness.light,
         child: Builder(
           builder: (context) {
+            setErrorBuilder(context);
```
andrzejchm, line 37: not really sure why we're setting a variable as part of the build method, and why this variable is static? Looks like flutter team at google was a bit drunk when implemented this O.o. :D Wouldn't it make more sense to have specialized widget that is responsible of building the Error UI ? I think `ErrorWidget` is more meant for general build errors and altering it's `builder` field you're updating the way all build errors are being displayed across the whole app.

starport_template/lib/stores/accounts_store.dart
```diff
@@ -202,6 +207,7 @@ class AccountsStore {
     return result.fold(
       (fail) {
         logError(fail);
+        errorDetails = fail.toString();
```
andrzejchm, line 210: we're loosing a lot of info here by calling `toString()`, couldn't we simply make errorDetails a, lets say `ErrorDetails` type that consists of the `dynamic error` and optional `StackTrace? stackTrace`.

starport_template/lib/starport_app.dart
Zfinix: https://docs.flutter.dev/testing/errors, yes they were drunk, i figured it would be great if we had our own custom error screen app-wide, gets rid of the red screen of death and allows us to still copy error messages with one click of a buttion

starport_template/lib/stores/accounts_store.dart
Zfinix: Will update this to use ErrorDetails

Zfinix: created CosmosErrorDetails

